PR #5, opened 2020-08-24
Minor updates

jimPruyne:
Just a few updates based on an editing pass


globus_automate_client/flows_client.py
```diff
@@ -160,7 +160,7 @@ def deploy_flow(
         available for execution on the Globus Automate Flows Service.
 
         :param flow_definition: A mapping corresponding to a Globus Flows
-            defintion.
+            definition.
```
urimandujano, line 163: I can't explain how many times I've spelled this word in the past two weeks and how many times I've missed the second "i"

docs/flow_creation_guide.md
```diff
@@ -1,34 +0,0 @@
-# Authoring Flows for the Globus Flows Service
```
urimandujano, line 1: I guess this is mostly out of date / captured in other parts of the docs?

urimandujano:
The edits look good and are welcome as they tie together the content across the pages better! Thanks for giving this another pass-through

PR #9, opened 2020-08-25
Supports a full ActionRequest via command line/SDK

urimandujano:
Adds support for creating the full ActionRequest document to the SDK and CLI. 

Previously, when launching an Action, the only parameters that could be set on it were `request_id` and `body`. The full Action Request spec outlines that a `POST` to `/run` may also contain `manage_by` and `monitor_by` fields. Not specifying these was fine but it meant that if someone ever wanted to set/use them, they'd be unable to via the client. This change corrects that. 

[ch2113]


jimPruyne:
Looks good. I kinda noticed this yesterday or so as well. Looks like this gets the job done nicely.

PR #11, opened 2020-09-01
An input schema for the multi-transfer flow example

jimPruyne:
A simple schema for this Flow. Can be specified when deploying using the --input-schema option on the CLI.


jimPruyne:
Agreed. It is good practice and it helps to make the Flows more self-documenting. This is probably a good first step in that direction.

PR #17, opened 2020-09-10
Allows special values in principal validation

urimandujano:
This fix essentially adds a second principal validator callback that accepts "public" and a third principal validator callback that accepts "all_authenticated_users" as valid values. These new callbacks use the old callback for validation if the principal is not one of the special values. [ch3716]

### DIFF
--- a/globus_automate_client/cli/actions.py
+++ b/globus_automate_client/cli/actions.py
@@ -7,7 +7,7 @@ import typer
 from globus_automate_client.action_client import create_action_client
 from globus_automate_client.cli.callbacks import (
     json_validator_callback,
-    principal_validator_callback,
+    principal_validator,
     url_validator_callback,
 )
 from globus_automate_client.cli.helpers import format_and_echo, verbosity_option
@@ -70,12 +70,12 @@ def action_run(
     manage_by: List[str] = typer.Option(
         None,
         help="A principal which may change the execution of the Action. [repeatable]",
-        callback=principal_validator_callback,
+        callback=principal_validator,
     ),
     monitor_by: List[str] = typer.Option(
         None,
         help="A principal which may view the state of the Action. [repeatable]",
-        callback=principal_validator_callback,
+        callback=principal_validator,
     ),
     verbose: bool = verbosity_option,
 ):
--- a/globus_automate_client/cli/actions.py
+++ b/globus_automate_client/cli/actions.py
@@ -65,7 +65,8 @@ def action_run(
         callback=json_validator_callback,
     ),
     request_id: str = typer.Option(
-        None, help=("An identifier to associate with this Action invocation request"),
+        None,
+        help=("An identifier to associate with this Action invocation request"),
     ),
     manage_by: List[str] = typer.Option(
         None,
--- a/globus_automate_client/cli/callbacks.py
+++ b/globus_automate_client/cli/callbacks.py
@@ -1,7 +1,7 @@
 import json
 import os
 import pathlib
-from typing import List, Optional
+from typing import AbstractSet, List
 from urllib.parse import urlparse
 from uuid import UUID
 
@@ -86,7 +86,7 @@ def text_validator_callback(message: str) -> str:
 
 
 def _base_principal_validator(
-    principals: List[str], *, special_vals=set()
+    principals: List[str], *, special_vals: AbstractSet[str] = frozenset()
 ) -> List[str]:
     """
     This validator ensures the principal IDs are valid UUIDs prefixed with valid
@@ -124,7 +124,7 @@ def principal_validator(principals: List[str]) -> List[str]:
     """
     A principal ID needs to be a valid UUID.
     """
-    return _base_principal_validator(principals, special_vals=None)
+    return _base_principal_validator(principals)
 
 
 def principal_or_all_authenticated_users_validator(principals: List[str]) -> List[str]:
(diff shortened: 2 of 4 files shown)


urimandujano:
@jimPruyne bump on this one as well

globus_automate_client/cli/callbacks.py
```diff
@@ -85,7 +85,7 @@ def text_validator_callback(message: str) -> str:
     return message
 
 
-def principal_validator_callback(principals: List[str]) -> List[str]:
+def principal_validator(principals: List[str]) -> List[str]:
```
jimPruyne, line 88: Can we just make something like `special_vals` a kwarg here that defaults to an empty set. Then, as we iterate check for presence in the `special_vals` set or having the proper principal id format here.

Then, the specialized functions can implement in terms of this by providing the kwarg.

globus_automate_client/cli/callbacks.py
```diff
@@ -114,6 +114,38 @@ def principal_validator_callback(principals: List[str]) -> List[str]:
...
 
 
+def principal_or_all_authenticated_users_validator(principals: List[str]) -> List[str]:
+    """
+    Certain fields expect values to be a valid Globus Auth UUID or one of a set
+    of special strings that are meaningful in the context of authentication.
+    This callback is a specialized form of the principal_validator where the
+    special value of 'all_authenticated_users' is accepted.
+    """
+    special_vals = {"all_authenticated_users"}
```
jimPruyne, line 124: per the above, this kinda becomes `return principal_validator(principals, special_vals={"all_authenticated_users"})`.

jimPruyne (changes requested):
Seems ok, but see comment for perhaps an idea for a bit less code and more generalizability in the future if needed.

urimandujano:
Good idea, Typer callbacks have some limitations which can be frustrating - I originally tried using a `functools.partial` to create functions mapped to specific `special_vals` but Typer didn't like when those functions were used as callbacks. The resulting code is smaller and easier to understand. Re-review?

globus_automate_client/cli/actions.py
```diff
@@ -65,7 +65,8 @@ def action_run(
         callback=json_validator_callback,
     ),
     request_id: str = typer.Option(
-        None, help=("An identifier to associate with this Action invocation request"),
+        None,
+        help=("An identifier to associate with this Action invocation request"),
```
jimPruyne, line 69: Not a complaint, but seems there's been some of this formatting stuff bouncing back and forth. I wonder if we somehow have different settings on black or if it is not consistent itself or whatever.

Perhaps something to be considered if we see this happening more. Definitely outside the scope of this PR.

globus_automate_client/cli/callbacks.py
```diff
@@ -85,15 +85,21 @@ def text_validator_callback(message: str) -> str:
     return message
 
 
-def principal_validator(principals: List[str]) -> List[str]:
+def _base_principal_validator(
+    principals: List[str], *, special_vals=set()
```
jimPruyne, line 89: I think using `set()` here is safe in terms of re-use. I guess something like: `special_vals: Set[str] = frozenset()` would be ultimately safe?

I'm just comparing this to general guidance where default values on kwargs like `[]` or `{}` for lists or dicts are not considered safe because, if I recall correctly, those values get created at interpret time and could get changed on multiple calls to the function.

globus_automate_client/cli/callbacks.py
```diff
@@ -114,20 +120,23 @@ def principal_validator(principals: List[str]) -> List[str]:
     return principals
 
 
+def principal_validator(principals: List[str]) -> List[str]:
+    """
+    A principal ID needs to be a valid UUID.
+    """
+    return _base_principal_validator(principals, special_vals=None)
```
jimPruyne, line 127: Could probably either omit `special_vals` or just pass `set()` here.

jimPruyne (commented):
I think I'm down to the nitpicky of nitpickiness here. One more look to see if you think the ideas here improve things any??

globus_automate_client/cli/actions.py
urimandujano: I noticed that too...  I've felt that VS code has been on the fritz lately and it might not be formatting correctly on my end. Will keep an eye out for when it happens again

globus_automate_client/cli/callbacks.py
urimandujano: That makes sense, I've convinced myself that using mutable values is fine bc they're used in the context of a CLI command but I think best practices should prevail. I think setting the default to `None` would be best in this case.

jimPruyne: I like frozenset. I've been using `Optional[CollectionType[base type]] = None` for so long, but recently occurred to me that there are cases where `Iterable[base type] = ()` for an empty tuple which is immutable or `Set[base type] = frozenset()` for sets is doable.

This isn't a silver bullet of course. `Iterable` isn't always sufficient (but maybe catches the old string is iterable problem when specified this way (the values returned from iterating a string aren't also strings that happen to be one character long, are they?)). And, I don't believe there's any form of a dict which is immutable, so there'll still probably be things that are `[Optinonal[Dict[...]] = None`, but maybe we apply this sort of thing when we can.

And, even in cases like this which are practically thinking probably safe, we start to train ourselves and get our code having some bit of consistency.

urimandujano: Consistency is a big win and probably worth re-wiring some of the assumptions that we bring into our code. I like the `Set[base type] = frozenset()` solution with maybe one change - `AbstractSet[str] = frozenset()`. The change has the benefit of silencing mypy warnings about whether a set or frozenset is used when the func gets called.

jimPruyne (approved):

PR #18, opened 2020-09-23
Adds input examples for running an action on the search delete AP

urimandujano:


examples/action_bodies/search_delete_by_subject.json
```diff
@@ -0,0 +1,5 @@
+{
+    "delete_by": "subject",
+    "search_index": "dea6af9a-cab8-4a2c-a8ba-5b7c73334d3b",
```
jimPruyne, line 3: Do you want a real (or real-ish) index id here, or a dummy one like the all 0's in the other example?

jimPruyne (changes requested):
The examples are super useful.

Can we also get an update to docs/source/globus_action_providers.rst?

examples/action_bodies/search_delete_by_subject.json
urimandujano: Good suggestion!

urimandujano:
Can't believe I forgot to update the main doc on what all our Action Providers are. Updated it with a description/usage with hopefully enough information for someone to get going with it.

jimPruyne (approved):

PR #19, opened 2020-09-23
Removes False-y values from action_run body

urimandujano:
jgaff reported that when using the SDK's `ActionClient` against the [now old] Datacite AP, he would get errors back indicating that  `monitor_by` and `manage_by` cannot be None. What happened is that the SDK would use `None` for `monitor_by` and `manage_by` if no value was supplied, which resulted in action-run requests being sent with those fields set to `None`.  At the AP, these action-run requests would be rejected bc the `ActionRequest` schema expects `monitor_by` and `manage_by` to be lists if they're set.

The fix is to simply remove False'y values from the action-run body before it get's sent to an AP. Because `monitor_by` and `manage_by` are optional, this will allow users to use the SDK's `ActionClient` without specifying values for those args.

### DIFF
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -84,6 +84,8 @@ class ActionClient(BaseClient):
             "monitor_by": monitor_by,
             "manage_by": manage_by,
         }
+        # Remove None / empty list items from the temp_body
+        body = {k: v for k, v in body.items() if v}
         return self.post(path, body)
 
     def status(self, action_id: str) -> GlobusHTTPResponse:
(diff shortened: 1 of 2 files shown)


globus_automate_client/action_client.py
```diff
@@ -84,6 +84,8 @@ def run(
             "monitor_by": monitor_by,
             "manage_by": manage_by,
         }
+        # Remove None / empty list items from the temp_body
+        body = {k: v for k, v in body.items() if v}
```
jimPruyne, line 88: But, what if `v` is actually the value `False`? Or what if it is the empty-string?

jimPruyne (changes requested):

urimandujano:
The proposed change look good. I hadn't thought about it but `request_id` and `body` could be empty/Falsey and the action should still go through. It's funny since I spent a while thinking about why "body" should be optional on the CLI and here I am making assumptions that it's not (: 

Let me validate that empty lists are ok before merging them

urimandujano:
Everything checks out! Good proposal

PR #21, opened 2020-09-29
Change schema for the Pass State

jimPruyne:
Make Parameters or InputPath ok, but also allow for neither by
using a slightly new version of the existing Parameters or InputPath
definition that does not include any `required` values.

[ch3823]

### DIFF
--- a/globus_automate_client/flows_schema.json
+++ b/globus_automate_client/flows_schema.json
@@ -154,6 +154,16 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


urimandujano (approved):
I was going to mention that `ParametersOrInputPath` only gets referenced by `Action` states now and if we expected that every Action would require input of some sort. I think the far more likely case is yes, an Action will require input and keeping it one-or-the-other required makes sense.

PR #22, opened 2020-09-30
Adds support for deleting-messages that have been processed

urimandujano:
When using the Queues service, a user needs to acknowledge (aka delete) that a message has been processed before the message is removed. If the message is not deleted, the Queue will place the message back at the head of the Queue where the message will once again be retrieved. At present, the CLI and SDK do not support the deletion action which makes it impossible for users to consume from a Queue to process work. This PR adds functionality at the SDK and CLI level supporting message deletion.

This PR also updates the help string on a Queue's `delivery_timeout` parameter. Essentially, this parameter controls how long the Queue waits for a message-deletion before assuming processing has failed and re-adding the message to the head of the Queue. There's work being done on the Queues service to allow `delivery_timeout` to be set to less than 60 seconds and default to 30.

[ch4008]

### DIFF
--- a/globus_automate_client/cli/queues.py
+++ b/globus_automate_client/cli/queues.py
@@ -162,7 +162,7 @@ def queue_delete(
     format_and_echo(queue, verbose=verbose)
 
 
-@app.command("receive")
+@app.command("message-receive")
 def queue_receive(
     queue_id: str = typer.Argument(...),
     max_messages: int = typer.Option(
@@ -179,7 +179,7 @@ def queue_receive(
     format_and_echo(queue, verbose=verbose)
 
 
-@app.command("send")
+@app.command("message-send")
 def queue_send(
     queue_id: str = typer.Argument(...),
     message: str = typer.Option(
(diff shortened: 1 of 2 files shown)


jimPruyne (commented):
Looks good overall. I had looked at this briefly, and what I couldn't decide for sure is how to name commands related to delete given that we have just `send` and `receive`.

- Should it be `delete-message` as here to avoid conflict with deleteing the queue?
- Should it be `queue message delete` and then alias or change others to `queue message {send, receive}`
- Should it be `queue message-{send, receive}`?

If either of the latter two, do we keep `queue {send, receive}` around for now for compatibility? I assume Globus Labs people are the only ones using queues right now, so we could probably get away with a breaking change at this point and let them know our reasoning.

urimandujano:
From my experience working using queues elsewhere it felt familiar to refer to the "delete" activity as an ack (I think that's how celery/rabbitmq refer to indicating a task was complete). Elsewhere in Python seen "task_done". My personal preference is for "ack" but it could get confusing if only done at the CLI level or if the CLI/SDK diverge from Queues terminology.

I think I'd like to keep a relatively flat subcommand space. `queue message-{send, receive, delete}` feels like the most natural extension.

jimPruyne (approved):

PR #24, opened 2020-11-14
Private_Properties should be Private_Parameters

jimPruyne:


urimandujano (approved):
Glad you made these edits to the doc! Glad they were found too - good job with the close reading @jonathon.

PR #26, opened 2020-12-02
Initial implementation for supporting user supplied authorizers

urimandujano:
- Decouples the SDK from the CLI via kwargs and callbacks
- create_*_client helper functions are removed from the SDK layer
- Added new_client classmethod to Client classes

### DIFF
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -1,16 +1,14 @@
 import uuid
-from typing import Any, Dict, Iterable, Mapping, Optional
+from typing import Any, Iterable, Mapping, Optional
 
-import requests
 from globus_sdk import (
     AccessTokenAuthorizer,
     ClientCredentialsAuthorizer,
     GlobusHTTPResponse,
     RefreshTokenAuthorizer,
 )
+from globus_sdk.authorizers.base import GlobusAuthorizer
 from globus_sdk.base import BaseClient
-
-from globus_automate_client.token_management import get_authorizer_for_scope
 
 
 class ActionClient(BaseClient):
@@ -44,11 +42,7 @@ class ActionClient(BaseClient):
         such as its expected ``action_scope``, its ``input_schema``, and who to
         contact when there's trouble.
         """
-        headers: Dict = {}
-        if self.authorizer is not None:
-            self.authorizer.set_authorization_header(headers)
-        resp = requests.get(self.base_url, headers=headers)
-        return self.default_response_class(resp, client=self)
+        return self.get("")
 
     def run(
         self,
@@ -118,54 +112,11 @@ class ActionClient(BaseClient):
         path = self.qjoin_path(action_id, "release")
         return self.post(path)
 
-
-def create_action_client(
-    action_url: str, action_scope: Optional[str] = None
-) -> ActionClient:
-    """
-    A helper function to handle creating a properly authenticated ``ActionClient``
-    which can operate against *Globus Action Provider Interface* compliant Action
-    Providers.
-
-    Given the ``action_url`` for a specific ActionProvider, this function will
-    attempt to create a valid ``ActionClient`` for interacting with that
-    ActionProvider. If the ``action_scope`` is not provided, this function will
-    attempt to discover the ``action_scope`` by querying the target Action
-    Provider's introspection endpoint. If the Action Provider is not configured
-    to allow public, unauthenticated access to its introspection endpoint, the
-    ``action_scope`` will be non-discoverable and authentication will fail.
-
-    With the ``action_scope`` available, the function will search for a valid
-    token using the fair_research_login library. In the event that tokens for
-    the scope cannot be loaded, an interactive login will be triggered. Once
-    tokens have been loaded, an Authorizer is created and used to instantiate
-    the ``ActionClient`` which can be used for operations against that Action
-    Provider.
-
-    :param action_url: The URL address at which the target Action Provider
-        exists
-    :param action_scope: The target Action Provider's Globus Auth Scope used
-        for authenticating access to it
-    """
-    if action_scope is None:
-        # We don't know the scope which makes it impossible to get a token,
-        # but create a client anyways in case this action provider is publicly
-        # visible without authentication.
-        temp_client = ActionClient(
-            "temp_client", base_url=action_url, app_name="tmp_client",
+    @classmethod
+    def new_client(cls, action_url: str, authorizer: GlobusAuthorizer):
+        return cls(
+            "action_client",
+            app_name="Globus Automate SDK - ActionClient",
+            base_url=action_url,
+            authorizer=authorizer,
         )
-        action_scope = temp_client.action_scope
-
-    if action_scope:
-        authorizer = get_authorizer_for_scope(action_scope)
-    else:
-        # Any attempts to use this authorizer will fail but there's nothing we
-        # can do without a scope.
-        authorizer = None
-
-    return ActionClient(
-        "action_client",
-        base_url=action_url,
-        app_name="action_client",
-        authorizer=authorizer,
-    )
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -7,7 +7,7 @@ from globus_sdk import (
     GlobusHTTPResponse,
     RefreshTokenAuthorizer,
 )
-from globus_sdk.authorizers.base import GlobusAuthorizer
+from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.base import BaseClient
 
 
@@ -113,7 +113,26 @@ class ActionClient(BaseClient):
         return self.post(path)
 
     @classmethod
-    def new_client(cls, action_url: str, authorizer: GlobusAuthorizer):
+    def new_client(
+        cls, action_url: str, authorizer: GlobusAuthorizer
+    ) -> "ActionClient":
+        """
+        Classmethod to simplify creating an ActionClient. Use this method when
+        attemping to create an ActionClient with pre-existing credentials or
+        authorizers.
+
+        :param action_url: The url at which the target Action Provider is
+            located.
+
+        :param authorizer: The authorizer to use for validating requests to the
+            Action Provider.
+
+        **Examples**
+            >>> authorizer = ...
+            >>> action_url = "https://actions.globus.org/hello_world"
+            >>> ac = ActionClient.new_client(action_url, authorizer)
+            >>> print(ac.run({"echo_string": "Hello from SDK"}))
+        """
         return cls(
             "action_client",
             app_name="Globus Automate SDK - ActionClient",
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -1,5 +1,5 @@
 import uuid
-from typing import Any, Iterable, Mapping, Optional
+from typing import Any, Iterable, Mapping, Optional, Type, TypeVar, Union
 
 from globus_sdk import (
     AccessTokenAuthorizer,
@@ -7,8 +7,9 @@ from globus_sdk import (
     GlobusHTTPResponse,
     RefreshTokenAuthorizer,
 )
-from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.base import BaseClient
+
+_ActionClient = TypeVar("_ActionClient", bound="ActionClient")
 
 
 class ActionClient(BaseClient):
@@ -17,6 +18,10 @@ class ActionClient(BaseClient):
         RefreshTokenAuthorizer,
         ClientCredentialsAuthorizer,
     )
+
+    AllowedAuthorizersType = Union[
+        AccessTokenAuthorizer, RefreshTokenAuthorizer, ClientCredentialsAuthorizer
+    ]
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
@@ -112,10 +117,26 @@ class ActionClient(BaseClient):
         path = self.qjoin_path(action_id, "release")
         return self.post(path)
 
+    def log(
+        self, action_id: str, limit: int = 10, reverse_order: bool = False
+    ) -> GlobusHTTPResponse:
+        """
+        Retrieve an Action's execution log history.
+
+        :param action_id: An identifier that uniquely identifies an Action
+            executed on this Action Provider.
+        :param limit: A integer specifying how many log records to return
+        :param reverse_order: Display the Action states in reverse-
+            chronological order
+        """
+        params = {"reverse_order": reverse_order, "limit": limit}
+        path = self.qjoin_path(action_id, "log")
+        return self.get(path, params=params)
+
     @classmethod
     def new_client(
-        cls, action_url: str, authorizer: GlobusAuthorizer
-    ) -> "ActionClient":
+        cls: Type[_ActionClient], action_url: str, authorizer: AllowedAuthorizersType
+    ) -> _ActionClient:
         """
         Classmethod to simplify creating an ActionClient. Use this method when
         attemping to create an ActionClient with pre-existing credentials or
--- a/globus_automate_client/client_helpers.py
+++ b/globus_automate_client/client_helpers.py
@@ -1,0 +1,82 @@
+from typing import Optional
+
+from globus_automate_client.action_client import ActionClient
+from globus_automate_client.flows_client import (
+    MANAGE_FLOWS_SCOPE,
+    PROD_FLOWS_BASE_URL,
+    FlowsClient,
+)
+from globus_automate_client.token_management import get_cli_authorizer
+
+
+def create_action_client(
+    action_url: str, action_scope: Optional[str] = None
+) -> ActionClient:
+    """
+    A helper function to handle creating a properly authenticated ``ActionClient``
+    which can operate against *Globus Action Provider Interface* compliant Action
+    Providers. This helper will look for and store tokens on a local filesystem,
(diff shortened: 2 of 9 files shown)


urimandujano:
@jimPruyne - still have some documentation/cleanup to do but this is what the implementation looks like. Each flow-action method now supports getting an authorizer via a kwarg but I also wanted to support the use case where there's a default way of looking up an authorizer via a callback. The CLI layer should function the same, SDK users may need to update their scripts depending on where they're importing things from/if they're fine using CLI level functions in their scripts.

globus_automate_client/action_client.py
```diff
@@ -1,17 +1,15 @@
...
 from globus_sdk import (
     AccessTokenAuthorizer,
     ClientCredentialsAuthorizer,
     GlobusHTTPResponse,
     RefreshTokenAuthorizer,
 )
+from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.base import BaseClient
 
-from globus_automate_client.token_management import get_authorizer_for_scope
```
jimPruyne, line 13: A good sign this is gone.

globus_automate_client/action_client.py
```diff
@@ -44,11 +42,7 @@ def introspect(self, **kwargs) -> GlobusHTTPResponse:
         such as its expected ``action_scope``, its ``input_schema``, and who to
         contact when there's trouble.
         """
-        headers: Dict = {}
-        if self.authorizer is not None:
-            self.authorizer.set_authorization_header(headers)
-        resp = requests.get(self.base_url, headers=headers)
-        return self.default_response_class(resp, client=self)
+        return self.get("")
```
jimPruyne, line 45: Also a good sign that this is gone. Should it be `self.get("/")`. It kinda reads better, but probably doesn't matter or perhaps even would cause a problem?

globus_automate_client/action_client.py
```diff
@@ -118,54 +112,30 @@ def release(self, action_id: str) -> GlobusHTTPResponse:
         path = self.qjoin_path(action_id, "release")
         return self.post(path)
 
+    @classmethod
+    def new_client(
+        cls, action_url: str, authorizer: GlobusAuthorizer
```
jimPruyne, line 117: I've been mucking about with a few ways to handle these factory things. I've tried:
```
T = TypeVar('T')
@classmethod
def new_thing(cls: Type[T]), ...) -> T:
```

And there's also 
```
from __future__ import annotations
class Foo:
@classmethod
def new_foo(cls, ...) -> Foo:
```

The latter is better and the from future isn't even necessary starting with something like 3.9, but I'm not sure it is even available for < 3.7 (I think we keep 3.6 as a valid version, but I'm not sure we've officially stated that).

So, I guess maybe try the former and especially see if in an editor like VSCode that uses these annotations well it accomplishes what we want in terms of type checking in use and completion and so on.

jimPruyne, line 117: Elsewhere we've stated:
```
    allowed_authorizer_types = (
        AccessTokenAuthorizer,
        RefreshTokenAuthorizer,
        ClientCredentialsAuthorizer,
    )
```
which is consumed by the `BaseClient`. If we can make the declaration of `authorizer` also support that, then we get the type system supporting what the run-time does. I think a `Union` would do the trick?

globus_automate_client/client_helpers.py
```diff
@@ -0,0 +1,82 @@
+from typing import Optional
+
+from globus_automate_client.action_client import ActionClient
+from globus_automate_client.flows_client import (
+    MANAGE_FLOWS_SCOPE,
+    PROD_FLOWS_BASE_URL,
+    FlowsClient,
+)
+from globus_automate_client.token_management import get_cli_authorizer
```
jimPruyne, line 9: Seems good having `cli` in the name to make this explicit.

globus_automate_client/client_helpers.py
```diff
@@ -0,0 +1,82 @@
...
+
+def cli_authorizer_callback(**kwargs):
+    flow_url = kwargs["flow_url"]
+    flow_scope = kwargs["flow_scope"]
+    client_id = kwargs["client_id"]
+    return get_cli_authorizer(flow_url, flow_scope, client_id)
+
+
+def create_flows_client(
+    client_id: str, base_url: str = PROD_FLOWS_BASE_URL
```
jimPruyne, line 56: Without looking around: does this still support the environment variable override?

globus_automate_client/client_helpers.py
```diff
@@ -0,0 +1,82 @@
...
+    ``FlowsClient`` which can operate against the Globus Automate Flows service.
+    This function will attempt to load tokens for the MANAGE_FLOWS_SCOPE using
+    the fair-research/native-login library. In the event that tokens for the
+    scope cannot be loaded, an interactive login Flow will be triggered. Once
+    tokens have been loaded, an Authorizer is created and used to instantiate
+    the ``FlowsClient``. To create a FlowsClient without referencing local
+    storage or triggering a login flow, instantiate an authorizer directly and
+    use the ``FlowsClient.new_client`` classmethod.
+
+    :param client_id: The Globus ID to associate with this instance of the
```
jimPruyne, line 69: Is there a reason we have `client_id` here but not on the similar action client helper?

globus_automate_client/flows_client.py
```diff
@@ -37,11 +37,13 @@
     RUN_STATUS_SCOPE,
 )
 
+AUTHORIZER_CALLBACK_TYPE = Callable[[Tuple[str, str, str]], GlobusAuthorizer]
```
jimPruyne, line 40: Maybe camel case with a starting capital to kinda match types (e.g. class names).

globus_automate_client/flows_client.py
```diff
@@ -194,7 +204,7 @@ def deploy_flow(
         """
         if validate_definition:
             validate_flow_definition(flow_definition)
-        self.authorizer = get_authorizer_for_scope(MANAGE_FLOWS_SCOPE, self.client_id)
+        self.authorizer = self.flow_management_authorizer
```
jimPruyne, line 207: This should just get set on the super class during `__init__` and thus we don't have to futz with setting `self.authorizer` any place else.

jimPruyne: Oops, replying to myself, maybe not. I'll have to think ahead. When we do `run` do we need a different scope on the Flows service or are we then using the scope on the Flow Id. Probably the latter, so maybe we do need this.

Hmm, need to think a bit here on how to make this as simple as possible. I guess the callback handles it ok, but in the perhaps common case where a user creates one `FlowClient` and then does a lookup and then a run sort of thing would be nice not to have to juggle authorizers or call the callback multiple times which may, in turn, cause the user to go through consent more than once.

I don't have the solution on this one right away...

globus_automate_client/flows_client.py
```diff
@@ -311,28 +320,55 @@ def list_flows(
...
+        return self.delete(f"/flows/{flow_id}", **kwargs)
+
+    def scope_for_flow(self, flow_id: str) -> str:
+        """
+        Returns the scope associated with a particular Flow
+
+        :param flow_id: The UUID identifying the Flow's scope to lookup
+        """
+        flow_url = f"{self.base_url}/flows/{flow_id}"
```
jimPruyne, line 344: I wonder if we can try a simple unauthenticated GET here first so that in the case (not sure how common) where the lookup can be done without authentication, we avoid having to make the user go through the consent flow or more specifically here give us back an authenticator.

globus_automate_client/flows_client.py
```diff
@@ -311,28 +320,55 @@ def list_flows(
...
-            ac = create_action_client(
-                f"{self.base_url}/flows/{flow_id}", VIEW_FLOWS_SCOPE
-            )
-            flow_scope = ac.action_scope
+            flow_scope = self.scope_for_flow(flow_id)
 
-        ac = create_action_client(f"{self.base_url}/flows/{flow_id}", flow_scope)
-        return ac
+        flow_url = f"{self.base_url}/flows/{flow_id}"
+        return self.get_authorizer_callback(
```
jimPruyne, line 367: Yeah, so this is the real point of a lot of this, and this looks good to me.

globus_automate_client/flows_client.py
```diff
@@ -349,7 +385,13 @@ def run_flow(
         :param flow_input: A Flow-specific dictionary specifying the input
             required for the Flow to run.
         """
-        ac = self._action_client_for_flow(flow_id, flow_scope)
+        if "authorizer" in kwargs:
```
jimPruyne, line 388: I think I like this too with the option of providing the kwarg. Not sure how this impacts my point above about a single authorizer set on the FlowsClient during __init__ so it doesn't have to get re-set elsewhere. Perhaps it ends up working?

globus_automate_client/flows_client.py
```diff
@@ -367,7 +409,13 @@ def flow_action_status(
         :param flow_action_id: The ID specifying the Action for which's status
             we want to query
         """
-        ac = self._action_client_for_flow(flow_id, flow_scope)
+        if "authorizer" in kwargs:
```
jimPruyne, line 412: maybe an internal helper for this to avoid duplicating the code. In said helper, can treat kwargs simply as a dict.

globus_automate_client/flows_client.py
```diff
@@ -478,49 +539,65 @@ def flow_action_log(
         :param reverse_order: An indicator for whether to retrieve the records
             in reverse-chronological order.
         """
-        if flow_scope is None:
-            flow_scope = self._scope_for_flow(flow_id)
-        self.authorizer = get_authorizer_for_scope(flow_scope, self.client_id)
+        if "authorizer" in kwargs:
```
jimPruyne, line 542: Seems that some of this is here because the `ActionClinet` is lacking the `log` method. That's a general Action method, so if it is not on the ActionClient it should move there and then this should follow the same pattern as the other Action related methods.

jimPruyne (commented):
I think this has hit the key points. Some cosmetic stuff and perhaps even a couple opportunities to simplify and DRY it a bit, but this feels like we've made a step in the right direction!

globus_automate_client/action_client.py
urimandujano: Ideally it wouldn't matter but I noticed the introspection endpoint in APT doesn't have a trailing "/" so making it `self.get("/")` results in 404s (which also means that introspecting an action with action-url's ending in "/" throws 404). I like the `self.get("/")` and will probably make the endpoint agnostic to trailing slashes.

globus_automate_client/client_helpers.py
urimandujano: I was sorely tempted to move the whole of `globus_automate_client.token_management` into the `cli` directory but opted not to to keep diffs manageable and to reduce the potential for breakage.

globus_automate_client/client_helpers.py
urimandujano: The env variable override happens at the CLI level so this function is blissfully unaware. This function kinda bridges the gap between the CLI and the SDK so if you see it worthwhile to add the env var override here I'd be fine with that.

globus_automate_client/client_helpers.py
urimandujano: Not really, but I'm not totally sure what `client_id` does in either scenario. Is it just used for identifying the application? Should we care if the CLI ID spills over and is used at the SDK?

globus_automate_client/flows_client.py
urimandujano: From my testing with this, it didn't look like we could introspect a Flow to pull it's scope without authentication: 
`
code | "AUTHENTICATION_ERROR"
detail | "Authorization header missing"
`
Is this supposed to be/ok to be public info? I was using the authorizer created with the MANAGE_SCOPES grant to do this and that seemed to be working fine (but adds some juggling of authorizers).

globus_automate_client/flows_client.py
urimandujano: From what I understand so far, the flow list/update/deploy/delete operations require consent to the MANAGE_FLOWS scopes, and the flow-run,flow-action-status,flow-action-release, flow-action-cancel need consents to Flow-specific scopes. And to auto-discover the Flow-specific scopes, you need to have consented to the MANAGE_FLOWS scopes so that in the end we have two authorizers with two different sets of consents.

So it sort of seems like there's two authorizers at play. In this implementation, the init method expects the one with the MANAGE_FLOWS consents, and the authorizer_callback is used to pull the flow-dependant authorizers. Is that mental model right? Maybe there's something simpler we could be doing?

globus_automate_client/flows_client.py
urimandujano: `list-flow-actions` is also something we could move into the ActionClient (as the action-enumerate op) but I'm not sure it's time for that.

globus_automate_client/action_client.py
urimandujano: Nevermind, it looks like this needs to be self.get("") to work with the Flows service as it exists. The best fix for this would be making our endpoints agnostic to trailing slashes. if we only do it in the CLI we have a lot of updates to do in docs/examples/code.

globus_automate_client/action_client.py
urimandujano: Added this - VSCode seems to reason well enough about the `new_client` return object with the `-> "FlowsClient` notation. I'm kinda impressed we naturally have this variety in editors bc it let's us test what works and doesn't work in a bunch of different editors.

urimandujano:
@jimPruyne I made some updates based on comments, you're welcome to take another pass when you can.

globus_automate_client/action_client.py
jimPruyne: Agreed, we should plan to make that change. Should be as easy as setting the property on the blueprint about trailing slashes.

globus_automate_client/action_client.py
jimPruyne: yeah, no need to change something that's ok. But, looking forward to not having to do that in the future.

globus_automate_client/client_helpers.py
jimPruyne: The client id is used to identify the native client requesting the grant from Auth. Auth will display the name of that client when it requests consent. I think this is ok here as I see some difference between action and flows client.

But, in the bit of context I see here, it is hard to think about, so I'm gonna look again at doing a review of the most recent commit and see what I see with more context.

globus_automate_client/flows_client.py
jimPruyne: I think you're probably right.

We should probably in the Flows service be looking at the value the user provided on `visible_to` and allow the unauthenticated case. But, that's a server-side problem first, then we can address here. So, probably leave as is for now.

globus_automate_client/flows_client.py
jimPruyne: I think you have it right. And, but this, all the operations on `MANAGE_FLOWS` is in the FlowsClient and we defer all the other operations to the ActionClient, so there's probably no need to juggle the authorizer within the FlowsClient (which is part of the goal of separating it this way).

globus_automate_client/flows_client.py
jimPruyne: Probably not time for that, but having the log in ActionClient is a good thing and I think I see that in the updated commit (again, will do a review on that separately).

globus_automate_client/action_client.py
```diff
@@ -1,15 +1,16 @@
...
 from globus_sdk import (
     AccessTokenAuthorizer,
     ClientCredentialsAuthorizer,
     GlobusHTTPResponse,
     RefreshTokenAuthorizer,
 )
-from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.base import BaseClient
 
+_ActionClient = TypeVar("_ActionClient", bound="ActionClient")
```
jimPruyne, line 12: Another usage I hadn't seen/used before. I think this better than having the string elsewhere.

Gotta love python, many ways to do everything...

globus_automate_client/action_client.py
```diff
@@ -18,6 +19,10 @@ class ActionClient(BaseClient):
         ClientCredentialsAuthorizer,
     )
 
+    AllowedAuthorizersType = Union[
```
jimPruyne, line 22: :+1:

globus_automate_client/action_client.py
```diff
@@ -112,10 +117,26 @@ def release(self, action_id: str) -> GlobusHTTPResponse:
         path = self.qjoin_path(action_id, "release")
         return self.post(path)
 
+    def log(
```
jimPruyne, line 120: yeah!!

globus_automate_client/client_helpers.py
```diff
@@ -6,11 +6,13 @@
...
 )
-from globus_automate_client.token_management import get_cli_authorizer
+from globus_automate_client.token_management import CLIENT_ID, get_cli_authorizer
 
 
 def create_action_client(
-    action_url: str, action_scope: Optional[str] = None
+    action_url: str,
+    action_scope: Optional[str] = None,
+    client_id: str = CLIENT_ID,
```
jimPruyne, line 15: This seems appropriate for the client helper layer. I think we have this right.

jimPruyne (commented):
I think this is looking good now.

I'd be inclined to Approve / merge, but before I do, let me know if you have any more questions/concerns.

urimandujano:
I'm good with the info the NativeClient ID makes sense to me so adding it to the helpers is good. A couple of action items to keep track of:

- make action provider toolkit tools not care about endpoints with trailing "/"
- make flows endpoints not care about trailing "/"
- potentially set "visible_to" on flows to be public

PR #28, opened 2020-12-08
A flow for performing a move operation

jimPruyne:
This adds an example of a flow that performs a move operation. We will
expect to deploy and maintain this flow for use by customers, but it
also provides an example of flow definition including an input
schema.

### DIFF
--- a/examples/flows/transfer-mv/definition.json
+++ b/examples/flows/transfer-mv/definition.json
@@ -3,6 +3,7 @@
   "StartAt": "Transfer",
   "States": {
     "Transfer": {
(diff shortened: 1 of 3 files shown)


urimandujano:
This looks really good and glad to see an example of a flow with error checking/handling tossed in. I wish there were a better way to narrate the state progression of the flow that didn't rely so much on the flow's top level comment. It may be something to figure out how to display on readthedocs.

A potential improvement here is adding an example `input.json` that validates against the schema.

jimPruyne:
@urimandujano Good suggestions. Was able to do some stuff to address each.

PR #31, opened 2020-12-23
Task/ch 4422 yaml support

jakeglobus:
Just looking for feedback on this point, particularly on any general style concerns or my overall approach to this fitting in with the code as it already exists. Please feel free to be critical!

### DIFF
--- a/examples/flows/hello-world-yaml/definition.yaml
+++ b/examples/flows/hello-world-yaml/definition.yaml
@@ -1,0 +1,14 @@
+Comment: Run the Hello World Action as the Flow's identity
+StartAt: RunHelloWorld
+States:
+  RunHelloWorld:
+    RunAs: Flow
+    Comment: Say Hello World
+    Type: Action
+    ActionUrl: 'https://actions.automate.globus.org/hello_world'
+    ActionScope: 'https://auth.globus.org/scopes/actions.globus.org/hello_world'
+    Parameters:
+      echo_string.$: $.input.echo_string
+      sleep_time.$: $.input.sleep
+    ResultPath: $.HelloResult
+    End: true
--- a/examples/flows/hello-world-yaml/input.yaml
+++ b/examples/flows/hello-world-yaml/input.yaml
@@ -1,0 +1,3 @@
+input:
+    echo_string: 'From automate cli'
+    sleep: 20
--- a/globus_automate_client/cli/callbacks.py
+++ b/globus_automate_client/cli/callbacks.py
@@ -1,6 +1,7 @@
 import json
 import os
 import pathlib
+import yaml
 from typing import AbstractSet, List
 from urllib.parse import urlparse
 from uuid import UUID
@@ -63,6 +64,42 @@ def json_validator_callback(body: str) -> str:
             raise typer.BadParameter(f"Invalid JSON: {e}")
     return body
 
+
+def yaml_validator_callback(body: str) -> str:
+    """
+    A user supplied body can be a properly formatted YAML string or a the name
+    of a file that contains valid YAML.  This validator ensures the user
+    supplies a valid YAML or valid filename containing valid YAML. Returns the
+    valid JSON string.
+    """
+    # Callbacks are run regardless of whether an option was explicitly set.
+    # Handle the scenario where the default value for an option is empty
+    if not body:
+        return body
+
+    # Reading from a file was indicated by prepending the filename with the @
+    # symbol -- for backwards compatability check if the symbol is present and
+    # remove it
+    if body.startswith("@"):
+        body = body[1:]
+
+    body_path = pathlib.Path(body)
+    yaml_body = None
+    if body_path.exists() and body_path.is_file():
+        with body_path.open() as f:
+            try:
+                yaml_body = yaml.safe_load(f)
+            except yaml.YAMLError as e:
+                raise typer.BadParameter(f"Invalid YAML: {e}")
+    elif body_path.exists() and body_path.is_dir():
+        raise typer.BadParameter("Expected file, received directory")
+    else:
+        try:
+            yaml_body = yaml.safe_load(body)
+        except yaml.YAMLError as e:
+            raise typer.BadParameter(f"Invalid YAML: {e}")
+
+    return json.dumps(yaml_body)
 
 def text_validator_callback(message: str) -> str:
     """
--- a/globus_automate_client/cli/callbacks.py
+++ b/globus_automate_client/cli/callbacks.py
@@ -2,11 +2,11 @@ import json
 import os
 import pathlib
 import yaml
+import typer
+
 from typing import AbstractSet, List
 from urllib.parse import urlparse
 from uuid import UUID
-
-import typer
 
 
 def url_validator_callback(url: str) -> str:
@@ -84,7 +84,7 @@ def yaml_validator_callback(body: str) -> str:
         body = body[1:]
 
     body_path = pathlib.Path(body)
-    yaml_body = None
+
     if body_path.exists() and body_path.is_file():
         with body_path.open() as f:
             try:
@@ -99,7 +99,13 @@ def yaml_validator_callback(body: str) -> str:
         except yaml.YAMLError as e:
             raise typer.BadParameter(f"Invalid YAML: {e}")
 
-    return json.dumps(yaml_body)
+    try:
+        yaml_to_json = json.dumps(yaml_body)
+    except TypeError as e:
+        raise typer.BadParameter(f"Unable to translate to JSON: {e}")
+
+    return yaml_to_json
+
 
 def text_validator_callback(message: str) -> str:
     """
@@ -192,3 +198,19 @@ def flows_endpoint_envvar_callback(default_value: str) -> str:
     defining the target Flow endpoint.
     """
     return os.getenv("GLOBUS_AUTOMATE_FLOWS_ENDPOINT", default_value)
+
+
+def input_validator_callback(value: str) -> str:
+    """
+    Allows for the input_format parameter to guide which input
+    validation scheme to use (JSON or YAML) with JSON as default
+    """
+    json_data = None
+
+    try:
+        json_data = json_validator_callback(value)
+    except:
+
+        json_data = yaml_validator_callback(value)
+    finally:
+        return json_data
--- a/globus_automate_client/cli/callbacks.py
+++ b/globus_automate_client/cli/callbacks.py
@@ -200,17 +200,27 @@ def flows_endpoint_envvar_callback(default_value: str) -> str:
     return os.getenv("GLOBUS_AUTOMATE_FLOWS_ENDPOINT", default_value)
 
 
-def input_validator_callback(value: str) -> str:
+def input_validator_callback(body: str) -> str:
     """
-    Allows for the input_format parameter to guide which input
-    validation scheme to use (JSON or YAML) with JSON as default
+    Checks if input is a file and loads it, otherwise
+    returns the body string passed in
     """
-    json_data = None
+    # Callbacks are run regardless of whether an option was explicitly set.
+    # Handle the scenario where the default value for an option is empty
+    if not body:
+        return body
 
-    try:
-        json_data = json_validator_callback(value)
-    except:
+    # Reading from a file was indicated by prepending the filename with the @
+    # symbol -- for backwards compatability check if the symbol is present and
+    # remove it
+    if body.startswith("@"):
+        body = body[1:]
 
-        json_data = yaml_validator_callback(value)
-    finally:
-        return json_data
+    body_path = pathlib.Path(body)
+    if body_path.exists() and body_path.is_file():
+        with body_path.open() as f:
+            body = f.read()
+    elif body_path.exists() and body_path.is_dir():
+        raise typer.BadParameter("Expected file, received directory")
+
+    return body
--- a/globus_automate_client/cli/callbacks.py
+++ b/globus_automate_client/cli/callbacks.py
@@ -1,12 +1,12 @@
 import json
 import os
 import pathlib
-import yaml
-import typer
-
 from typing import AbstractSet, List
 from urllib.parse import urlparse
 from uuid import UUID
+
+import typer
+import yaml
 
 
 def url_validator_callback(url: str) -> str:
@@ -43,8 +43,7 @@ def json_validator_callback(body: str) -> str:
     # Reading from a file was indicated by prepending the filename with the @
     # symbol -- for backwards compatability check if the symbol is present and
     # remove it
-    if body.startswith("@"):
-        body = body[1:]
+    body = body.lstrip("@")
 
     body_path = pathlib.Path(body)
     if body_path.exists() and body_path.is_file():
@@ -80,8 +79,8 @@ def yaml_validator_callback(body: str) -> str:
     # Reading from a file was indicated by prepending the filename with the @
     # symbol -- for backwards compatability check if the symbol is present and
     # remove it
-    if body.startswith("@"):
-        body = body[1:]
+    body = body.lstrip("@")
+
 
     body_path = pathlib.Path(body)
 
@@ -213,8 +212,8 @@ def input_validator_callback(body: str) -> str:
     # Reading from a file was indicated by prepending the filename with the @
     # symbol -- for backwards compatability check if the symbol is present and
     # remove it
-    if body.startswith("@"):
-        body = body[1:]
+    body = body.lstrip("@")
+
 
     body_path = pathlib.Path(body)
     if body_path.exists() and body_path.is_file():
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -56,6 +56,10 @@ class ActionStatus(str, Enum):
     active = "ACTIVE"
     inactive = "INACTIVE"
 
+class FlowInputFormat(str, Enum):
+    json = "json"
+    yaml = "yaml"
+
 
 app = typer.Typer(short_help="Manage Globus Automate Flows")
 
@@ -76,8 +80,8 @@ def flow_deploy(
     definition: str = typer.Option(
         ...,
         help=(
-            "JSON representation of the Flow to deploy. May be provided as a filename "
-            "or a raw JSON string."
+            "JSON or YAML representation of the Flow to deploy. May be provided as a filename "
+            "or a raw string representing a JSON object or YAML definition."
         ),
         prompt=True,
         callback=json_validator_callback,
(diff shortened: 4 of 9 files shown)


examples/flows/hello-world-yaml/definition.yaml
```diff
@@ -0,0 +1,14 @@
...
+StartAt: RunHelloWorld
+States:
+  RunHelloWorld:
+    RunAs: Flow
+    Comment: Say Hello World
+    Type: Action
+    ActionUrl: 'https://actions.automate.globus.org/hello_world'
+    ActionScope: 'https://auth.globus.org/scopes/actions.globus.org/hello_world'
+    Parameters:
+      echo_string.$: $.input.echo_string
```
jimPruyne, line 11: Not a comment on this PR, just on the proposed approach of using YAML, the mixing of JSON path into the YAML feels slightly weird to me, but I'm not proposing anything different unless a) someone knows of a YAML equivalent to JSON path and 2) it can easily be translated to JSON Path. Given that this is referencing the Flow state, and the Flow state's "native" representation is JSON, I doubt this is even a good idea (introducing some hypothetical YAML-style path syntax), but I thought I'd mention to see if it elicits any thoughts.

examples/flows/hello-world-yaml/definition.yaml
```diff
@@ -0,0 +1,14 @@
+Comment: Run the Hello World Action as the Flow's identity
```
jimPruyne, line 1: I agree with @urimandujano 's comment on the call. This does read more cleanly than the equivalent JSON to me.  Seems to validate the general idea and makes the rest of the effort worth it.

If nothing else, the LoC count goes down compared to the JSON equivalent which is not a bad thing.

examples/flows/hello-world-yaml/input.yaml
```diff
@@ -0,0 +1,3 @@
+input:
```
jimPruyne, line 1: good to see the input provided in addition to the flow definition.

globus_automate_client/cli/callbacks.py
```diff
@@ -1,12 +1,13 @@
 import json
 import os
 import pathlib
+import yaml
+import typer
```
jimPruyne, line 5: I'm not sure what triggered the move of the typer import. Ideally,we're running on a consistent config for isort and that's doing the job. Given that typer is alphabetically before yaml, I'm guessing isort wasn't run here.

@urimandujano Perhaps you can comment on isort config, running lint steps etc.

But, not a problem at all at this draft stage.

globus_automate_client/cli/flows.py
```diff
@@ -57,6 +60,30 @@ class ActionStatus(str, Enum):
...
+    json = "json"
+    yaml = "yaml"
+
+
+def _process_definition(definition: str, input_format: str) -> Mapping[str, Any]:
+    """
+    Turn input strings into dicts per input format type (FlowInputFormat)
+    """
+    flow_dict = None
+    if input_format == FlowInputFormat.json:
```
rudyardrichter, line 73: There was a whole discussion on the philosophy of `Enum` in #devchat back in November, from which, to be honest, I'm not certain what the main conclusions were. That said, I would argue for this sort of usage:
```suggestion
    if FlowInputFormat(input_format) is FlowInputFormat.json:
```
which has the benefits of
- Generating an informative `ValueError` if `input_format` is somehow not in the enum
- Making a comparison that's within the universe of your enum type rather than comparing string values, which is a "value" comparison rather than a "type" comparison, to be really picky about it

To that effect, I might even argue for changing the signature to
```python
_process_definition(definition: str, input_format: FlowInputFormat)
```
and here using
```suggestion
    if input_format is FlowInputFormat.json:
```
while calling `_process_definition(definition, FlowInputFormat(some_string_value))` elsewhere.

Also roping in @urimandujano and @jimPruyne for this discussion, sorry to blow up this one bit of code @jakeglobus but felt it's worth us having this discussion as it relates our codebases and this is a decent example.

globus_automate_client/cli/callbacks.py
```diff
@@ -64,6 +65,48 @@ def json_validator_callback(body: str) -> str:
...
+        body = body[1:]
+
+    body_path = pathlib.Path(body)
+
+    if body_path.exists() and body_path.is_file():
+        with body_path.open() as f:
+            try:
+                yaml_body = yaml.safe_load(f)
+            except yaml.YAMLError as e:
+                raise typer.BadParameter(f"Invalid YAML: {e}")
```
jimPruyne, line 93: Would be curious what the error output from the exception looks like. If there's anything in the exception state that is helpful in tracking down the bad syntax (e.g. line/column numbers) AND if those are not presented nicely in the output from `{e}` it might be worth doing some custom formatting of the error message.

globus_automate_client/cli/helpers.py
```diff
@@ -9,7 +10,11 @@
 )
 
 
-def format_and_echo(result: Union[GlobusHTTPResponse, str], verbose=False) -> None:
+def format_and_echo(
+    result: Union[GlobusHTTPResponse, str],
+    dumper: Callable = lambda result, **args: json.dumps(result, indent=4, sort_keys=True),
```
rudyardrichter, line 15: After staring at it for a second, I don't see an actual issue with this, but my python senses would be tingling less if this was an `Optional[Callable]` defaulting to None and updated to this lambda in the body if None. Or if there was a small `dump_to_json` function
```python
def dump_to_json(result, *args, **kwargs):
    return json.dumps(result, indent=4, sort_keys=True)
```
and this defaults to `dumper: Callable = dump_to_json`

globus_automate_client/cli/callbacks.py
```diff
@@ -155,3 +198,29 @@ def flows_endpoint_envvar_callback(default_value: str) -> str:
     defining the target Flow endpoint.
     """
     return os.getenv("GLOBUS_AUTOMATE_FLOWS_ENDPOINT", default_value)
+
+
+def input_validator_callback(body: str) -> str:
```
jimPruyne, line 203: Still getting through the whole thing, so this comment may become moot or more clear, but seems redundant with the file loading stuff in the yaml validator.

Perhaps I'll discover this structure going forward, but it does seem like doing a layer first of determining whether the parameter loads as a file first and returning that file content or the original string, followed by doing appropriate json/yaml parsing is a good approach. Thus, we don't have file open/read in multiple places.

I guess doing so means that all the json/yaml parsing will be string rather than file based, but assuming that doesn't mean we lose information that would be helpful in error reporting, that seems ok to me.

rudyardrichter (commented):
- Appreciate your comprehensive use of `typing`!

globus_automate_client/cli/callbacks.py
urimandujano, line 5: I think there's some back and forth on the sort order for imports going on. It looks like you're using vscode, have you configured `isort` on save/manually? We have some `isort` config in pyproject.toml and the `isort` command should pick the config up from there

examples/flows/hello-world-yaml/definition.yaml
urimandujano, line 1: NICE - I like that we can start sharing examples for how Flow definitions in yaml look like. I for one think it helps with readability a lot

globus_automate_client/cli/flows.py
```diff
@@ -76,11 +103,11 @@ def flow_deploy(
...
         ...,
         help=(
-            "JSON representation of the Flow to deploy. May be provided as a filename "
-            "or a raw JSON string."
+            "JSON or YAML representation of the Flow to deploy. May be provided as a filename "
+            "or a raw string representing a JSON object or YAML definition."
         ),
         prompt=True,
-        callback=json_validator_callback,
+        callback=input_validator_callback,
```
jimPruyne, line 110: Looks like this is headed toward what I was thinking above, this'll do the file management stuff and then the parsing can assume it has the string representation.

Probably means the file open/reading stuff can be removed from the json/yaml-specific validators?

globus_automate_client/cli/flows.py
```diff
@@ -345,7 +419,12 @@ def flow_list(
     flows = fc.list_flows(
         roles=[r.value for r in roles], marker=marker, per_page=per_page
     )
-    format_and_echo(flows, verbose=verbose)
+
+    format_and_echo(
+        flows,
+        json.dumps if output_format == FlowDisplayFormat.json else yaml.dump,
```
jimPruyne, line 425: I'm fine with this method of determining which dump function to use or the if/else combination used above. May want to standardize on one or the other, though.

globus_automate_client/cli/flows.py
```diff
@@ -57,6 +60,30 @@ class ActionStatus(str, Enum):
...
+            flow_dict = json.loads(definition)
+        except json.JSONDecodeError as e:
+            raise typer.BadParameter(f"Invalid JSON: {e}")
+    elif input_format == FlowInputFormat.yaml:
+        try:
+            flow_dict = yaml.safe_load(definition)
+        except Exception as e:
+            raise typer.BadParameter(f"Invalid YAML: {e}")
+
+    return flow_dict
```
urimandujano, line 84: I think it may be easier to take an input string and just try loading it as JSON first - if it triggers a `JSONDecodeError`, we can try loading it as YAML - and if that still triggers an `Exception`, we can raise a `typer.BadParameter` exception. This way, we remove the burden of needing to indicate the input format from the user.

globus_automate_client/cli/helpers.py
jimPruyne, line 15: I liked the callable as the parameter when I read it above, this is a particularly cute approach with the lambda. Makes me wonder in other places where `json.dumps` is specifically provided if it should be removed to get this behavior, or if this lambda, and perhaps a corresponding one with indent etc. set for yaml should be defined more globally and re-used on all the calls to this function.

I guess if such lambdas or other function definitions are defined they could be in this same source file and imported into the caller.

globus_automate_client/cli/flows.py
```diff
@@ -161,7 +198,12 @@ def flow_deploy(
         input_schema_dict,
         validate_definition=validate,
     )
-    format_and_echo(result, verbose=verbose)
+
+    # Match up output format with input format
+    if input_format == FlowInputFormat.json:
+        format_and_echo(result, json.dumps, verbose=verbose)
+    elif input_format == FlowInputFormat.yaml:
+        format_and_echo(result, yaml.dump, verbose=verbose)
```
urimandujano, line 206: I'm inclined to add the format detection into the `format_and_echo` function to help keep the logic simple at this top level. Similar to my previous comment, I think a trying to dump json then backing off to dumping as yaml may make the most sense. Since `result` will always be a GlobusHTTP response, we can treat it as we need to in `format_and_echo`.

globus_automate_client/cli/helpers.py
```diff
@@ -21,7 +26,9 @@ def format_and_echo(result: Union[GlobusHTTPResponse, str], verbose=False) -> No
         result = result.data
     else:
         color = typer.colors.GREEN
-    typer.secho(json.dumps(result, indent=4, sort_keys=True), fg=color)
+
+    typer.secho(dumper(result, indent=4, sort_keys=True), fg=color)
```
jimPruyne, line 30: Passing the indent/sort_keys to dumper seems a bit dangerous unless we know all possible dumper functions can take those as args. Perhaps the re-usable function/lambda that can be used by callers is a cleaner approach?

globus_automate_client/cli/helpers.py
urimandujano, line 15: I like this a lot, it paves the way for the day we want to support JSON/YAML output on the CLI

globus_automate_client/cli/main.py
```diff
@@ -6,6 +6,12 @@
 from globus_automate_client.cli import actions, flows, queues
 from globus_automate_client.token_management import CONFIG_PATH
 
+# Monkey patching out the unsafe load capability
+# Only use safe_load for our purposes
+import yaml
+del yaml.unsafe_load
```
jimPruyne, line 12: Never saw this approach before. I like the ~paranoid~ defensive coding approach.

globus_automate_client/cli/callbacks.py
```diff
@@ -155,3 +198,29 @@ def flows_endpoint_envvar_callback(default_value: str) -> str:
...
+        body = body[1:]
+
+    body_path = pathlib.Path(body)
+    if body_path.exists() and body_path.is_file():
+        with body_path.open() as f:
+            body = f.read()
+    elif body_path.exists() and body_path.is_dir():
+        raise typer.BadParameter("Expected file, received directory")
+
+    return body
```
urimandujano, line 226: Can we generalize this? I think the current implementation only supports loading JSON, but if we want to support YAML everywhere, we'd need to be able to parse input as YAML too.

Now that I think about it, I think we'd essentially want to use a single function to handle reading user input. The function could then delegate to the `yaml_validator_callback` and `json_validator_callback` to load the input.

globus_automate_client/cli/callbacks.py
```diff
@@ -64,6 +65,48 @@ def json_validator_callback(body: str) -> str:
...
+            yaml_body = yaml.safe_load(body)
+        except yaml.YAMLError as e:
+            raise typer.BadParameter(f"Invalid YAML: {e}")
+
+    try:
+        yaml_to_json = json.dumps(yaml_body)
+    except TypeError as e:
+        raise typer.BadParameter(f"Unable to translate to JSON: {e}")
+
+    return yaml_to_json
```
urimandujano, line 107: This is a good catch - I think we want to standardize all input as it enters the CLI layer and standardizing on json is a good thing.

jimPruyne (commented):
I think this is looking pretty good. As may be expected at this stage, opportunities for consistency and removing some duplicate code in places, but I feel good about the overall approach.

We are on the same page I think from what we agreed on in that selecting yaml as a format has to be done via the input argument and we're not attempting the parse-and-guess approach.

Some not so important stylistic stuff: In clubhouse we have the types feature, bug, chore and so I (we?) typically name branches like `feature/`, `bug/`, `chore/` and a useful name. I've used issue ids / numbers in branch names in the past, and I've never decided whether I like it or not. We should be putting the ch story id in the body of the PR description in the form `[chNNN]` as that supports automation within clubhouse.

I also have the personal preference for reducing the number of commits in the PR though again that's mostly just personal preference. I tend to do a lot of interactive rebase operations before pushing. That's another thing that can still be done later if desired given we're at draft stage.

So, looking good. The upshot of the result I do think will be a win for users though we'll also need to make sure we add the docs so people know it is there as well.

In terms of the docs, I'm fine putting it in this branch/PR or doing a separate one when we know that this has settled/merged. When that's done, also we can plan a release to pypi/readthedocs.

globus_automate_client/cli/callbacks.py
```diff
@@ -64,6 +65,48 @@ def json_validator_callback(body: str) -> str:
...
+    # Callbacks are run regardless of whether an option was explicitly set.
+    # Handle the scenario where the default value for an option is empty
+    if not body:
+        return body
+
+    # Reading from a file was indicated by prepending the filename with the @
+    # symbol -- for backwards compatability check if the symbol is present and
+    # remove it
+    if body.startswith("@"):
+        body = body[1:]
```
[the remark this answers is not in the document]
urimandujano: I like this!

globus_automate_client/cli/flows.py
urimandujano: I remember that :) I'm not sure it's as valuable to generate an informative `ValueError` here since typer will try to do that itself when it parses the CLI argument. But the verdict's still out on other places we use Enums. I do like the `is` optional for readability though.

examples/flows/hello-world-yaml/definition.yaml
urimandujano: Just leaving a comment to say the mixing is sorta lame but I don't know of any YAML-path. If I squint real hard I can sorta mentally map the path onto a json input but I can't say it'll work in every scenario.

globus_automate_client/cli/callbacks.py
urimandujano: vscode's isort integration isn't super well thought out. Since we don't have CI here it amounts to configuring vscode to run isort on save everytime OR running it manually. I added this in my vscodes `settings.json`:

`    "[python]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": true
        }
    },`

globus_automate_client/cli/flows.py
jimPruyne: @urimandujano Yeah, just to be clear, @jakeglobus  and I talked about this. We might end up going that way in due time, but for first cut, we decided to make it a user-set option in this first pass.

urimandujano: Ah I didn't realize there'd been a discussion about the approach. Then that sounds good and the implementation looks like a good one!

globus_automate_client/cli/callbacks.py
jakeglobus: This was sorta my attempt at generalizing the input validation. So far our input seems to be limited to two types: string or file. I do think we could have some better error handling here for handling the None body and maybe some other cases I haven't thought of. Something to think about I guess

globus_automate_client/cli/callbacks.py
jakeglobus: Here is a screenshot of trying to parse invalid yaml:
Invalid yaml parse

globus_automate_client/cli/flows.py
jakeglobus: Yes this is what I was thinking. Parse the input (string or file) then return a string to be parsed as the appropriate type (json, yaml)

globus_automate_client/cli/flows.py
jakeglobus: I'm glad you saw this, it reminded me to ask you about what to do here. So this code handles the output format without using an output format param, it just matches up the output to be the same type as the input. This makes sense to me but I guess it boils down to these two choices:
1. Use the same output format as output format
2. Add an output format that defaults to json (same as elsewhere)

What do you think?

examples/flows/hello-world-yaml/definition.yaml
jakeglobus: So just to clarify, since YAML is a superset of JSON, the JSON path IS the YAML path. So it's not technically mixing anything but by supporting YAML we are introducing some confusion and I can see how that could cause problems. Not sure what the best thing to do is besides just document it really well. I'm open to any solution

globus_automate_client/cli/flows.py
urimandujano: Sorry this may be coming a little late - I've seen other CLI that let you specify output format via a CLI switch. We want to get there at some point (especially as we eventually look to move into the main globus cli) and I think it's reasonable to only support JSON output until we're there.

For reference the main globus CLI has these output formats: `-F, --format [unix|json|text]`

PR #33, opened 2021-01-11
Task/ch 4422 yaml support

jakeglobus:
Adding support for YAML input and output

https://app.clubhouse.io/globus/story/4422/ahm-client-accept-yaml-input-provide-yaml-output


globus_automate_client/cli/actions.py
```diff
@@ -14,6 +16,45 @@
...
+    action_dict = None
+    if input_format is ActionInputFormat.json:
+        try:
+            action_dict = json.loads(body)
+        except json.JSONDecodeError as e:
+            raise typer.BadParameter(f"Invalid JSON: {e}")
+    elif input_format is ActionInputFormat.yaml:
+        try:
+            action_dict = yaml.safe_load(body)
+        except Exception as e:
```
jimPruyne, line 52: Is this the most precise Exception type that can be specified for `yaml.safe_load`?

Also, it seems like the main complication of using the same `DUMPERS` type definition for a "LOADER" usage is the possible exception types, right? One could do LOADERS where the value on the dict is the loader function and a tuple of exception types if one were so inclined. 

The LOADER bit is pretty hypothetical, doesn't really matter to me, but if we can be more specific on the exception type that seems good. If that's just what yaml.safe_load returns, just say so and I'm good here too.

globus_automate_client/cli/actions.py
```diff
@@ -14,6 +16,45 @@
 app = typer.Typer(short_help="Manage Globus Automate Actions")
 
 
+class ActionInputFormat(str, Enum):
+    json = "json"
+    yaml = "yaml"
```
jimPruyne, line 21: This is going to look out of order, but perhaps it will make more sense after seeing the comments related to DUMPERS and hypothetical LOADERS. Seems one could simply put the `load` method on this InputFormat Enum and DUMPERS on the OutputFormat Enum. Still could make use of the Dict approach to map to the real dump functions and just make that a class property.

Again, splitting hairs to some degree. I'm ok with any of the mentioned approaches including what's already in the code.

globus_automate_client/cli/actions.py
```diff
@@ -28,14 +69,22 @@ def action_introspect(
...
+    ),
 ):
     """
     Introspect an Action Provider's schema.
     """
     ac = create_action_client(action_url, action_scope)
     if ac is not None:
         result = ac.introspect()
-        format_and_echo(result, verbose=verbose)
+        format_and_echo(result, DUMPERS[output_format], verbose=verbose)
```
jimPruyne, line 87: So, if `dump` were moved into `ActionOutputFormat` this could become `output_format.dump(result)` maybe?

globus_automate_client/cli/callbacks.py
```diff
@@ -64,6 +64,47 @@ def json_validator_callback(body: str) -> str:
...
+    """
+    # Callbacks are run regardless of whether an option was explicitly set.
+    # Handle the scenario where the default value for an option is empty
+    if not body:
+        return body
+
+    # Reading from a file was indicated by prepending the filename with the @
+    # symbol -- for backwards compatability check if the symbol is present
+    # remove it if present
+    body = body.lstrip("@")
```
jimPruyne, line 82: Does this and all the file handling still need to be here, or has the input callback insured that by the time this is called, file loading is complete? Sorry if I've lost sight of the overall flow within the diff.

globus_automate_client/cli/flows.py
```diff
@@ -57,6 +60,32 @@ class ActionStatus(str, Enum):
     inactive = "INACTIVE"
 
 
+class FlowInputFormat(str, Enum):
```
jimPruyne, line 63: This seems to clear up something that was confusing me on the use of `Action` part of `ActionInputFormat`. Do we need separate Input or Output format enums for Actions and Flows? I'd think not, so perhaps we move the Input and Output format enums to a common place and just call them `InputFormat` and `OutputFormat`.

globus_automate_client/cli/flows.py
```diff
@@ -139,14 +168,24 @@ def flow_deploy(
...
     Deploy a new Flow.
     """
     fc = create_flows_client(CLIENT_ID, flows_endpoint)
-    flow_dict = json.loads(definition)
+
+    flow_dict = _process_definition(definition, input_format)
+
     if input_schema is not None:
-        input_schema_dict = json.loads(input_schema)
+        input_schema_dict = yaml.safe_load(input_schema)
```
jimPruyne, line 188: This seems to assume that the input schema is yaml. Probably need to check the input format flag here.

I guess a single input format flag means one couldn't specify, for example, the definition in json and the schema in yaml (or vice versa), but I'm good with that.

globus_automate_client/cli/flows.py
```diff
@@ -233,14 +277,22 @@ def flow_update(
...
 ):
     """
     Update a Flow.
     """
     fc = create_flows_client(CLIENT_ID, flows_endpoint)
-    flow_dict = json.loads(definition)
+    flow_dict = _process_definition(definition, input_format)
     if input_schema is not None:
-        input_schema_dict = json.loads(input_schema)
+        input_schema_dict = yaml.safe_load(input_schema)
```
jimPruyne, line 295: same as previous on assuming yaml?

globus_automate_client/cli/flows.py
```diff
@@ -424,17 +502,25 @@ def flow_run(
...
+        show_default=True,
+    ),
 ):
     """
     Run an instance of a Flow. The argument provides the initial state of the Flow.
     """
     fc = create_flows_client(CLIENT_ID, flows_endpoint)
     if flow_input is not None:
-        flow_input_dict = json.loads(flow_input)
+        flow_input_dict = yaml.safe_load(flow_input)
```
jimPruyne, line 519: another yaml assumption?

globus_automate_client/cli/flows.py
```diff
@@ -665,7 +751,7 @@ def flow_action_log(
         if output_format == FlowDisplayFormat.graphviz:
             typer.echo(graphviz_out.source)
         else:
-            graphviz_out.render(f"flows-output/graph", view=True, cleanup=True)
+            graphviz_out.render("flows-output/graph", view=True, cleanup=True)
```
jimPruyne, line 754: good catch on the unneeded f-sttring

jimPruyne (changes requested):
Want to double check if there are places with yaml assumptions.

Other things are stylistic, but unifying the Input/Output format enums so Actions and Flows are not separate seems to make sense to me.

Of course, let me know if I'm missing something...

examples/flows/hello-world-yaml/input-schema.yaml
```diff
@@ -0,0 +1,14 @@
+additionalProperties: false
```
urimandujano, line 1: It's nice that we have a full example for a Flow's components defined in YAML. Once this is advertised better, I'm curious if folks stick with JSON or move to YAML

examples/flows/transfer-mv-yaml/definition.yaml
```diff
@@ -0,0 +1,81 @@
+Comment: >-
```
urimandujano, line 1: It's extra nice to see the right chomping bit, I think using it in the example will save a lot of users headaches.

globus_automate_client/cli/callbacks.py
```diff
@@ -224,3 +221,44 @@ def input_validator_callback(body: str) -> str:
         raise typer.BadParameter("Expected file, received directory")
 
     return body
+
+
+def flow_input_validator(body: str) -> str:
```
urimandujano, line 226: Nice catch that plain JSON can get loaded by the YAML loader. Does this mean we can remove the other more specific json/yaml loaders to do something more generic as is here?

globus_automate_client/cli/callbacks.py
jakeglobus: Both yaml_validator_callback and json_validator_callback are not used anywhere. Since we are handling it in a different way now, would it be preferable to remove them? I personally tend towards removing code that is no longer used but I understand there are different approaches to handling this.

globus_automate_client/cli/flows.py
jakeglobus: Right now we do probably need at least separate output format enums, since flows can be displayed as 4 different formats while actions can only be displayed as two. In the future I would suggest creating ouput and input helpers and centralizing all the code and logic within. In applications like this I think it's helpful to channel input and output through single entry and exit points within the code.

jakeglobus: The input formats could likely be consolidated since they are the same two formats (json and yaml)

globus_automate_client/cli/flows.py
jakeglobus: When I wrote this I was assuming that we would like to be able to accommodate the definition in json and the schema in yaml or vice versa. I can update these spots to be more specific and throw an error when the format does not match.

In this case, yaml can load json just fine, so it's not assuming it's just using the superset to parse the input_schema regardless of the format.

globus_automate_client/cli/callbacks.py
jakeglobus: I think it's okay to remove both of the yaml and json validator callbacks and my first instinct is to remove them, but I wasn't sure how you all handled no longer used code. Sometimes people preserve it for future use, but I'm not sure that's necessary here

globus_automate_client/cli/actions.py
jakeglobus: That was an oversight, I'll update it to use the proper YAML exception yaml.YAMLError. Good catch, thanks!

globus_automate_client/cli/callbacks.py
urimandujano: I feel similarly. If we need to we can return to git history.

globus_automate_client/cli/flows.py
urimandujano: Consolidating input formats seems like a good approach. Given the amount of enums we're using (shout out @typer), it may make sense to split the enums out into their own `enum.py` file. We may be able to get away with something clever by having an `InputFormat` and `OutputFormat` and then a separate FlowsOutputFormat that can take the extra values (+graphviz and image). I'm also fine with avoiding subclassing and creating a separate output format enum for Flows.

jakeglobus: Okay with my last commit I think I've addressed these concerns (at least partially) and provided a patter for going forward with the different Enum types and helper function interactions.

globus_automate_client/cli/flows.py
```diff
@@ -161,7 +191,12 @@ def flow_deploy(
         input_schema_dict,
         validate_definition=validate,
     )
-    format_and_echo(result, verbose=verbose)
+
+    # Match up output format with input format
```
jimPruyne, line 195: That makes sense. Good idea.

globus_automate_client/cli/flows.py
```diff
@@ -287,24 +332,36 @@ def flow_lint(
...
     try:
         if validate:
             validate_flow_definition(flow_dict)
     except FlowValidationError as fve:
         typer.secho(str(fve), fg=typer.colors.RED)
         raise typer.Exit(code=1)
+
     graph = graphviz_format(flow_dict)
-    if output_format == FlowDisplayFormat.json:
+    if output_format is FlowDisplayFormat.json:
```
jimPruyne, line 357: good one

jimPruyne (approved):
I think this is looking good now. I'd say lets go ahead as is now.

jimPruyne:
@urimandujano let us know if you're going to take another pass through this.

Else, @jakeglobus this is looking good to me.

urimandujano (approved):
This looks ready to me! LGTM

PR #37, opened 2021-01-29
process_input should not try to load arguments that are None

jakeglobus:
Found a bug I introduced while deploying the secure egress prototype.

### DIFF
--- a/globus_automate_client/cli/helpers.py
+++ b/globus_automate_client/cli/helpers.py
@@ -45,16 +45,16 @@ def display_http_details(response: GlobusHTTPResponse) -> None:
 
 
 def process_input(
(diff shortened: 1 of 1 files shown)


globus_automate_client/cli/helpers.py
```diff
@@ -51,6 +51,10 @@ def process_input(
     Turn input strings into dicts per input format type (InputFormat)
     """
     input_dict = None
+
+    if input_arg is None:
```
urimandujano, line 55: Since this has potential to be `None`, can we update the type annotation to reflect that? Also, do we need to create the `input_dict` at the top of the func?

jakeglobus: Good points, just pushed the changes

urimandujano (approved):

PR #39, opened 2021-02-04
Add label argument to flow run

rudyardrichter:
Would appreciate a pointer to where in the docs this should be included (if at all—I couldn't find anything similar with `--format` or `--input`)

### DIFF
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -507,6 +507,12 @@ def flow_run(
         case_sensitive=False,
         show_default=True,
     ),
+    label: str = typer.Option(
+        None,
+        "--label",
+        "-l",
+        help="Optional label to mark this run.",
+    )
 ):
     """
     Run an instance of a Flow. The argument provides the initial state of the Flow.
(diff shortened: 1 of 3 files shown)


globus_automate_client/cli/flows.py
```diff
@@ -507,14 +507,20 @@ def flow_run(
         case_sensitive=False,
         show_default=True,
     ),
+    label: str = typer.Option(
```
urimandujano, line 510: Is there a max length that this label can or should be? Transfer caps at 128 but I'm not sure what we're doing on Flows.

urimandujano (commented):
This looks good! Can you add the `label` option on the CLI for actions too to keep the interfaces into actions/flows in sync?

For CLI help I think we've asked folks to run `--help` on the various subcommands. We don't have any docs that formally document all the options (at least not yet, it just hasn't been an ask yet).

globus_automate_client/cli/flows.py
rudyardrichter: This crossed my mind on the [API side](https://github.com/globusonline/flows/pull/137) too. 128 sounds like a reasonable limit. In this case I think I'd rather include that on only the API side? since the API already has to also validate the input body. If that makes sense I'll add it to the API PR.

rudyardrichter:
Ok, thanks @urimandujano. I'll add label for actions.

globus_automate_client/cli/flows.py
rudyardrichter: [Added length check to that PR](https://github.com/globusonline/flows/pull/137/files#diff-ac504e479ec42f07e22c65fc287e2206cdfb5f63dd20f9eb88764ba9eb66efdeR474-R483)

urimandujano (approved):
Thanks for the updates!

PR #38, opened 2021-02-02
Adds exception handling to the CLI layer

urimandujano:
- Ignores exceptions thrown when trying to auto-introspect an AP.  The errors will be explicitly thrown when trying to use the created `ActionClient`
- Supports `GlobusAPIErrors` in printing functions
- Request details now print to `stderr` so that tools like `jq` can parse command output when the `verbosity` flag is set

PR #40, opened 2021-02-05
Fixes filename too long errors when reading JSON directly from the command line input

urimandujano:
When reading input from the user, the CLI determines if the input is a file _first_ by doing filesystem checks. If it is a valid file, we load the file's contents; if it is not a valid file, we use the input as the content directly. When providing JSON directly on the command line, it is very easy to create a long-enough JSON string that triggers a filename too long error. This PR essentially loads the input as content whenever a filename too long error occurs.

This PR builds off the [Adds exception handling to the CLI layer](https://github.com/globus/globus-automate-client/pull/38) PR

### DIFF
--- a/globus_automate_client/cli/actions.py
+++ b/globus_automate_client/cli/actions.py
@@ -4,6 +4,7 @@ from typing import List
 
 import typer
 import yaml
+from globus_sdk import GlobusAPIError
 
 from globus_automate_client.cli.callbacks import (
     input_validator_callback,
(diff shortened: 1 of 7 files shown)


urimandujano:
Opening this up for review now since I'm hoping to get this merged by Weds. To see the changes specific to this PR, it'll help to narrow the diffs to what's in the 2nd commit.

jimPruyne (approved):
Glad you pointed out this builds on the CLI error handling one. I think I identified the changes specific to this PR and it looks good. Sensible ordering of the file/parsing checks it seems.

PR #41, opened 2021-02-10
Remove ActionScope from example flows

jimPruyne:
Including the scope in the flow definition is an anti-pattern at this
point because introspection by the Flows service does a better job.

Also, seems my editor settings have changed since last I visited these
files and indent size has shrunk, likely making for any ugly
diff. Sorry, but I guess since I prefer the 2 space indent it is
better in the long run.


urimandujano (approved):
This seems to match up well with our docs where we say that `ActionScope` is optional. The less fields in our state, the better.

PR #42, opened 2021-02-10
Add dry run parameter to flow run and deploy

rudyardrichter:


jimPruyne:
@rudyardrichter is this PR ready for review (in spite of the merge conflict)?

rudyardrichter:
@jimPruyne it is now, yes

globus_automate_client/flows_client.py
```diff
@@ -413,7 +422,7 @@ def run_flow(
         authorizer = self._get_authorizer_for_flow(flow_id, flow_scope, kwargs)
         flow_url = f"{self.base_url}/flows/{flow_id}"
         ac = ActionClient.new_client(flow_url, authorizer)
-        return ac.run(flow_input, **kwargs)
+        return ac.run(flow_input, dry_run=dry_run, **kwargs)
```
urimandujano, line 425: I'm still not sure that I like making dry_run a field on `ActionClients` in general. How would we feel about doing a manual request/response (without the ActionClient) if doing a dry_run?

```
self.authorizer = authorizer
return self.post(f"{self.base_url}/flows/{flow_id}/dry_run", params=params, **kwargs)
```
I haven't tested this so I'm not sure if it works.

globus_automate_client/cli/flows.py
```diff
@@ -170,6 +170,14 @@ def flow_deploy(
         case_sensitive=False,
         show_default=True,
     ),
+    dry_run: bool = typer.Option(
+        False,
+        "--dry-run",
+        help=(
+            "Do a dry run of deploying the flow to test your definition without"
+            " actually making changes."
```
urimandujano, line 178: I think it would be useful to add more information about what this is validating, beyond the info a user would get just trying to deploy a Flow like normal.

globus_automate_client/cli/flows.py
```diff
@@ -507,14 +516,22 @@ def flow_run(
         case_sensitive=False,
         show_default=True,
     ),
+    dry_run: bool = typer.Option(
+        False,
+        "--dry-run",
+        help=(
+            "Do a dry run with your input to this flow to test the input without"
+            " actually running anything."
```
urimandujano, line 524: `dry_run` in the case of running a Flow checks validation against a Flow schema? Or is it checking more than that? Not sure what creating an `Action` on the Flow's side is validating.

globus_automate_client/flows_client.py
rudyardrichter: I would be fine with making it a separate method but I think one way or another it belongs on the client.

I'd argue for the current setup making sense in that, now with the URL-based differentiation of dry runs, the requests to `/flows` or `/flows/dry-run` are identical, so using the same method with a flag makes sense to avoid duplication.

urimandujano: Right I think it should be on the `FlowClient` for sure, I'm wondering if adding it as an option on the `ActionClient` is potentially misleading (since I imagine 99% of `ActionProviders` wouldn't support a dry-run type of endpoint). I'd be willing to sacrifice some duplication for the sake of keeping the `kwargs` on `ActionClient.run` accurate.

rudyardrichter: Yeah, I get the concern on Action Providers not generically supporting this. Though with the flow run/action equivalency this is probably going to look weird in one place or another as the first(?) point of divergence there, at least implementation-wise. I guess I would summarize my current stance as, I agree with your objections in principle, but since the clients _in the client codebase_ are for us and not the users, I would rather keep it as is (easiest/smallest code footprint) and not let users try to `--dry-run` an action for the time being.

What would be your take on changing `run_flow` so that it's pointing the action client directly at the dry run URL? Maybe that's a more elegant compromise.

urimandujano: The client's in the codebase are what get exposed as part of the SDK, which is why I have some concerns about adding `dry_run` onto the `Action.run` interface. We could document that `dry_run` only makes sense if the target Action is the Flows service, but that asks the questions of where to document that and how to handle the case where the flag is set on ActionProviders that don't implement a `dry_run` endpoint. It seems simpler to me to avoid the potential for any confusion by special-casing the implementation of dry-run on the FlowsClient.

rudyardrichter: Ohh. My mistake, I was assuming the CLI client was separate somehow to the SDK. Sorry to have wasted some of your time over that misunderstanding. Clearly I haven't spent enough time on the SDK side. :slightly_smiling_face: 

To be more specific about my previous suggestion, how would you feel about this for the flow run method:
```python
        authorizer = self._get_authorizer_for_flow(flow_id, flow_scope, kwargs)                                                                                               
        flow_url = f"{self.base_url}/flows/{flow_id}"
        if dry_run:
            flow_url += "/dry-run"                                                                             
        ac = ActionClient.new_client(flow_url, authorizer)                                                                                       
        return ac.run(flow_input, dry_run=dry_run, **kwargs)
```
and leaving the action run method alone?

urimandujano: IIRC the `run` method will make a request to the base URL + `/run` so I think in this case the URL would be off, but I forget what the dry-run endpoint on the Flow's side would be. If there's a good way to overload the client so that `run()` hits the `/dry-run`endpoint I'd be for that.

globus_automate_client/cli/flows.py
urimandujano: @rudyardrichter can you elaborate a little on what `--dry-run` gives feedback on? I still feel like we could be clearer about it in the help text.

urimandujano:
@rudyardrichter can you take care of the merge conflict? I think my only other ask is that the `dry-run` help options on the CLI have their text updated with more info on what the dry-run is doing.

jimPruyne (approved):

jimPruyne:
@rudyardrichter @urimandujano I went ahead and attempted a fix for the merge conflicts here. But, I haven't tested or anything. So, it is at least worth a re-review of the code and possibly a checkout and test to make sure things work as expected.

jimPruyne:
@rudyardrichter @urimandujano Anything still waiting on this PR?

jimPruyne:
@rudyardrichter @urimandujano One more bump on this so we could get it off the CH board. Seems a conflict has arisen in the meantime...

PR #44, opened 2021-02-19
Export flow validation function

jimPruyne:
Make the flow validation function visible at the top level, and provide the default CLIENT_ID for flow client creation


urimandujano (approved):
Loooks really good! Thanks for handling the feedback and implementing!

PR #46, opened 2021-03-08
Adds Filtering and Ordering on Flow List and Flow Action List

urimandujano:
Adds support in the CLI and SDK for the new filtering/ordering options available on the Flow List and Flow Action List endpoints and it's corresponding usage documentation. 

In the SDK, the general approach is to accept a series of filters/ordering options as dictionaries which get mapped to query parameters sent off to the Flows service.

In the CLI, I added fairly generic `--filter` and `--orderby` which accept `key=value` formatted strings. I think this is the most generalizable way of supporting adhoc options as they change or get updated on the server. There's still the issue of where a user should look to find which fields support filtering and ordering. I think I'd rather keep that information out of the CLI for now and bank on some Flows documentation providing that information.

### DIFF
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -395,8 +395,8 @@ def flow_list(
         "--filter",
         help="A filtering criteria in the form 'key=value' to apply to the "
         "resulting Flow listing. The key indicates the filter, the value "
-        "indicates the pattern to match. Multiple patterns may for a single key "
-        "may be specified as a comma seperated string, the results for which will "
+        "indicates the pattern to match. Multiple patterns for a single key may "
+        "be specified as a comma seperated string, the results for which will "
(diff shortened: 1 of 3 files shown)


globus_automate_client/cli/flows.py
```diff
@@ -596,6 +596,25 @@ def flow_actions_list(
         min=1,
         max=50,
     ),
+    filters: Optional[List[str]] = typer.Option(
+        None,
+        "--filter",
+        help="A filtering criteria in the form 'key=value' to apply to the "
+        "resulting Action listing. The key indicates the filter, the value "
+        "indicates the pattern to match. Multiple patterns may for a single key "
```
jakeglobus, line 604: I think maybe this should be "...may match for a a single key "

jimPruyne (approved):

PR #49, opened 2021-03-26
Add HTTP Timeouts

urimandujano:
Adds configurable 10 second http timeouts to the FlowsClient and ActionClients. In practice, this will take 20 seconds before failing when used via the CLI -> 10 seconds to try to connect to an Action/Flow to introspect the scope, and another 10 seconds to try to run the operation.


jimPruyne (approved):
This is a good idea. Esp now that our backend also has a timeout on calling to the APs, this makes the client and the backend more consistent, so AP devs in particular will get further coerced to making their stuff perform.

PR #51, opened 2021-04-04
Adds documentation around monitor_by and manage_by

urimandujano:
I started updating the docs but realized it would be more effort than I thought it would be. This PR address some concerns around what exactly an acceptable principal value is, both in the `CLI` and `SDK` levels for `ActionClients` and `FlowsClients`. It also explicitly adds the `manage_by` and `monitor_by` arguments to the `FlowClient.run()` method for better visibility (they used to just be lumped into `kwargs` and passed onto the underlying `ActionClient.run()` call).

Also:
- Fixes bugs in live rendering where Globus Auth logins weren't displaying correctly when watching an executing Action.
- Updates `flow lint` output to clearly indicate the linting results, instead of outputting Flow definitions.
- Adds support to `flow display` for visualizing local or remote Flow definitions.
- Removes faulty visualization options on flow_list

### DIFF
--- a/docs/source/quick_start.rst
+++ b/docs/source/quick_start.rst
@@ -46,7 +46,7 @@ SDK in a Python script:
 
 Use of the CLI and SDK to invoke any services requires authentication. Upon
 first interaction with any Action Provider or the Flows Service via the CLI, a
-text prompt will appear directing the user to a web URL where that can proceed
+text prompt will appear directing the user to a web URL where they can proceed
 through an authentication process using Globus Auth to consent to the service
 they are using the CLI to interact with. This typically only needs to be done
 once, the first time a particular service is invoked. Subsequently, the cached
--- a/docs/source/quick_start.rst
+++ b/docs/source/quick_start.rst
@@ -45,16 +45,16 @@ SDK in a Python script:
     off with the CLI interface.
 
 Use of the CLI and SDK to invoke any services requires authentication. Upon
-first interaction with any Action Provider or the Flows Service, a
-browser window will be opened to proceed through an Authentication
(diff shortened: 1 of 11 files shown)


docs/source/quick_start.rst
```diff
@@ -45,16 +45,16 @@ SDK in a Python script:
     off with the CLI interface.
 
 Use of the CLI and SDK to invoke any services requires authentication. Upon
-first interaction with any Action Provider or the Flows Service, a
-browser window will be opened to proceed through an Authentication
-process using Globus Auth to consent for use by the CLI with the
-service it is interacting with. This typically only needs to be done
+first interaction with any Action Provider or the Flows Service via the CLI, a
+text prompt will appear directing the user to a web URL where that can proceed
```
jakeglobus, line 49: think this is a typo, should be "where they can proceed"

globus_automate_client/cli/flows.py
```diff
@@ -458,16 +447,29 @@ def flow_display(
...
-    try:
-        flow_get = fc.get_flow(flow_id)
-    except GlobusAPIError as err:
-        format_and_echo(err, verbose=verbose)
+    if not flow_definition and not flow_id:
+        raise typer.BadParameter("Either FLOW_ID or --flow_definition should be set.")
+    if flow_definition and flow_id:
+        raise typer.BadParameter(
+            "Only one of FLOW_ID or --flow_definition should be set."
+        )
```
jakeglobus, line 459: nice, always good to provide more validation and user messages about what went wrong

jakeglobus (approved):
Looks good to me outside of that one typo

PR #54, opened 2021-04-22
Various bugfixes + output formatting updates

urimandujano:
- Adds more separation between verbose and regular output
- Colorizes verbose output separately
- Outputs verbose output to stderr
- Omits orderings qarg if not set
- Defaults optional args in list_flow_actions to None.

### DIFF
--- a/globus_automate_client/flows_client.py
+++ b/globus_automate_client/flows_client.py
@@ -382,7 +382,7 @@ class FlowsClient(BaseClient):
             params["per_page"] = str(per_page)
         if filters is not None:
             params.update(filters)
-        if orderings is not None:
+        if orderings:
             builder = []
             for field, value in orderings.items():
                 builder.append(f"{field} {value}")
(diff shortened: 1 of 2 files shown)


globus_automate_client/flows_client.py
```diff
@@ -645,7 +643,10 @@ def list_flow_actions(
                 builder.append(f"{field} {value}")
             params["orderby"] = ",".join(builder)
 
-        return self.get(f"/flows/{flow_id}/actions", params=params, **kwargs)
+        self.authorizer = self._get_authorizer_for_flow(flow_id, flow_scope, kwargs)
+        response = self.get(f"/flows/{flow_id}/actions", params=params, **kwargs)
+        self.authorizer = self.flow_management_authorizer
```
jakeglobus, line 648: Is this because the authorizer for the flow is different than the authorizer for managing flows?

jakeglobus (commented):
Sorry it took me so long to get to this!

globus_automate_client/flows_client.py
urimandujano: Yup, the Globus SDK client's can only support a single `authorizer` at a time, and the `authorizer` is used to authenticate the requests (`self.get`). Since we need 1 `authorizer` to handle interacting with the Flows service, and another one for dealing with individual Flows we end up juggling them around like this in the client operations. Setting `self.authorizer` to be `flow_management_authorizer` sets the Flows client up to be used to hit the Flow's service again.

jakeglobus (approved):

PR #55, opened 2021-05-02
Updates to support multiple environments

jimPruyne:


globus_automate_client/flows_client.py
```diff
@@ -155,6 +167,11 @@ def validate_flow_definition(flow_definition: Mapping[str, Any]) -> None:
     return
 
 
+def _get_flows_base_url_for_environment():
+    environ = os.environ.get("GLOBUS_SDK_ENVIRONMENT")
+    return _ENVIRONMENT_FLOWS_BASE_URLS[environ]
```
urimandujano, line 172: Can we be extra defensive here by doing a `get` that defaults to one of the production URL keys?

globus_automate_client/flows_client.py
```diff
@@ -700,7 +717,7 @@ def new_client(
         client_id: str,
         authorizer_callback: AuthorizerCallbackType,
         authorizer: AllowedAuthorizersType,
-        base_url: str = PROD_FLOWS_BASE_URL,
+        base_url: Optional[str] = None,
```
urimandujano, line 720: If we do the `get` above we can deal with a plain `str` type default and not worry about users passing in None, empty strings, or worse...

examples/flows/transfer-mv-yaml/definition.yaml
```diff
@@ -8,7 +8,7 @@ States:
       Run the initial transfer operation from the source ep/source path to the
       destination ep/destination path
     Type: Action
-    ActionUrl: 'https://actions.globus.org/transfer/transfer'
+    ActionUrl: 'https://prod.actions.automate.globus.org/transfer/transfer'
```
urimandujano, line 11: I don't know how complicated this would be but I think keeping our service URLs short and env agnostic would be good UX. As far as our users are concerned, they should only ever need prod and a URL with prod in it is unnecessary info. Maybe we can talk about what we'd need to do to keep using the `https://actions.globus.org/transfer/transfer` style urls

globus_automate_client/flows_client.py
jimPruyne: I was defensive by making `None` one of the keys in `_ENVIRONMENT_FLOWS_BASE_URL` Perhaps that is a bit farther a way so harder to see, but it also seems defensive in a decent way? I'm fine changing if you have a preference.

jimPruyne: Ah, read the next comment, and it triggered what I think you were saying: users could provide any old value in their environment and that might fault here.

I guess that begs the question whether an unknown value should default to prod or whether it should throw an error. I guess if I was going for `sandbox` and typo-d it, I'd rather have an error than actually going to prod. But, probably a cleaner error than an arbitrary `KeyError`.

globus_automate_client/flows_client.py
jimPruyne: I want a passed in value to override anything based on the `GLOBUS_SDK_ENVIRONMENT`. So, I went to the optional str so that nothing passed will known that using that env var is the next level, and then default from that is our usual prod path.

Does that make sense now that I've explicitly listed the precedence (and does that precedence make sense)?

examples/flows/transfer-mv-yaml/definition.yaml
jimPruyne: Hmm, I don't know where this change came from. I agree that it shouldn't be here, and bad on me for not seeing it in the commit diffs.

globus_automate_client/cli/flows.py
```diff
@@ -84,6 +79,12 @@ def _process_flow_input(flow_input: str, input_format) -> Mapping[str, Any]:
 
 app = typer.Typer(short_help="Manage Globus Automate Flows")
 
+_flows_env_var_option = typer.Option(
```
urimandujano, line 82: Thanks for this, setting the same envvar option on every command was starting to grind my gears. At this point I'd probably be fine looking for it once in `os.environ` directly but we can save that for a future refactor.

globus_automate_client/flows_client.py
urimandujano: Ah that makes sense. I hadn't thought of that and I think that actually falls in line with the main SDK's behavior. That's a good detail to catch!

urimandujano (approved):
This PR has good timing - I'd just run into issues using the CLI for different environments and ended up having to blast away my tokens file. 

To the question about how to handle misconfiguration - my time working on schema validation has given me some appreciation for helpful errors earlier rather than silent misbehaviors later. I feel good about what's here, thanks for being proactive with this need!

PR #66, opened 2021-05-12
Flows List default roles are Flows API default roles

urimandujano:
It's happened often enough where CLI users are confused why they get different Flow output on the Flow List command. This PR removes the default role filtering (used to default to created_by) so that if the CLI command is run without any role filters, the results returned are dependent on the API.

### DIFF
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -364,7 +364,7 @@ def flow_lint(
 @app.command("list")
 def flow_list(
     roles: List[FlowRole] = typer.Option(
-        [FlowRole.created_by],
+        [],
         "--role",
         "-r",
         help="Display Flows where you have the selected role. [repeatable]",


jimPruyne:
I kind of thought the created_by was a sane default. Perhaps this can be addressed via documentation on the option rather than a change of the default? Given the server side default change the list could be pretty large? Perhaps we need to discuss with the team before deciding on this?

urimandujano:
For folks not versed in our various Flow roles I think this leads to confusion (at least it's the source of many questions) when they use the CLI vs the WebApp. This lowers the barrier of entry and I think letting the server decide what gets returned makes it easier to document too. In this case I feel like the server defaults (minding performance) is worth getting on the same page about. Discussion is always welcome.

jimPruyne (approved):

rudyardrichter (approved):

PR #68, opened 2021-05-13
Cleanup logout/revoke multi-environment

jimPruyne:
Cleans up the handling of the logout and revoke operations in multiple
environments. Each of these will only change the tokens stored for the
environment set by GLOBUS_SDK_ENVIRONMENT.

### DIFF
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -1,10 +1,11 @@
+import copy
 import json
 import os
 import pathlib
 import platform
 import sys
 from json import JSONDecodeError
-from typing import Any, Dict, List, NamedTuple, Optional, Set, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Set, Union
 
 import click
 import typer
@@ -57,7 +58,7 @@ TokensInTokenCache = Dict[str, Union[TokenSet, Dict[str, TokenSet]]]
 
 
 class TokenCache:
-    def __init__(self, token_store: str):
+    def __init__(self, token_store: Union[pathlib.Path, str]):
         self.token_store = token_store
         self.tokens: TokensInTokenCache = {}
         self.modified = False
@@ -68,7 +69,7 @@ class TokenCache:
         We will sub-key the full token set for environments other than production
         """
         environ = os.environ.get("GLOBUS_SDK_ENVIRONMENT")
-        if environ in {None, "production", "prod"}:
+        if environ in {None, "production", "prod", "default"}:
             return self.tokens
         environ_cache_key = "__" + environ
         if environ_cache_key not in self.tokens:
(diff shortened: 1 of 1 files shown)


globus_automate_client/cli/auth.py
```diff
@@ -279,7 +300,7 @@ def refresh_handler(
 def get_authorizer_for_scope(
     scope: str, client_id: str = CLIENT_ID
 ) -> Optional[GlobusAuthorizer]:
-    authorizers = get_authorizers_for_scopes([scope], client_id=CLIENT_ID)
+    authorizers = get_authorizers_for_scopes([scope], client_id=client_id)
```
urimandujano, line 303: Good catch!

globus_automate_client/cli/auth.py
```diff
@@ -161,6 +162,26 @@ def default(x):
...
+        environment_aware: bool = True,
+        callback: Optional[Callable[[str, TokenSet], bool]] = None,
+    ) -> None:
+        if environment_aware:
+            tokens = self.tokens_for_environment
+        else:
+            tokens = self.tokens
+        for scope, token_set in copy.copy(tokens).items():
+            if scope.startswith("__"):
+                continue
```
urimandujano, line 176: Is there a certain format for the token storage? Would you be able to write a docstring / just describe what the expected format is? Mainly for the key since.

globus_automate_client/cli/auth.py
```diff
@@ -295,32 +316,34 @@ def get_access_token_for_scope(scope: str) -> Optional[str]:
...
-    try:
-        os.remove(token_store)
-    except OSError:
-        click.echo("couldn't remove token cache file", err=True)
-        return False
+def logout(token_store: Union[pathlib.Path, str] = DEFAULT_TOKEN_FILE) -> bool:
+    cache = TokenCache(token_store)
+    cache.load_tokens()
+    cache.clear_tokens()
+    cache.save_tokens()
```
urimandujano, line 323: I really like how this reads

globus_automate_client/cli/auth.py
```diff
@@ -295,32 +316,34 @@ def get_access_token_for_scope(scope: str) -> Optional[str]:
...
     if not client:
         click.echo("failed to get auth client", err=True)
         return False
     cache = TokenCache(token_store)
     cache.load_tokens()
-    if not logout(token_store):
-        return False
-    for token_set in cache.tokens.values():
+
+    def revoker(scope: str, token_set: TokenSet) -> bool:
```
urimandujano, line 335: I guess I could make a case that the revoke should be a method on the `TokenSet` itself (or on the token store) and the client would be a parameter to the operation, since presumable the `TokenSet` would know how to revoke its tokens. That being said I can also see how it's nice to keep the `TokenSet` kinda dumb and keep the logic elsewhere. Up to you if you want to move the revoke operation into the class.

urimandujano:
@jimPruyne I made a few comments and suggestions so consider this ready for another pass.

globus_automate_client/cli/auth.py
```diff
@@ -161,6 +162,26 @@ def default(x):
...
+        if environment_aware:
+            tokens = self.tokens_for_environment
+        else:
+            tokens = self.tokens
+        for scope, token_set in copy.copy(tokens).items():
+            if scope.startswith("__"):
+                continue
+            token_set: TokenSet = token_set  # type checking stuff
+            do_remove = True
+            if callback is not None:
```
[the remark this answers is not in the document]
jimPruyne: Even if callback is None, we still remove the tokens from the tokenset. Adding the callback just lets the caller process the removed token as it happens. So, I think this is good as is.

globus_automate_client/cli/auth.py
jimPruyne: I added a bit of doc and a constant to hopefully clearup the use of the `"__"` string.

globus_automate_client/cli/auth.py
jimPruyne: Yeah, that's a good point. Also a good point about keeping it dumb. I think of it as really managing the persistent store more than having any real understanding of Auth. I guess that breaks in that it has the update from the oauth response thing, so maybe that should be generic-ized to some other update sort of function, but I'll keep it as is for now.

jimPruyne:
@urimandujano Responded to comments and pushed a change

urimandujano:
All looks good thanks!

PR #70, opened 2021-05-14
Link out to webapp in examples

rudyardrichter:


docs/source/authoring_flows.rst
```diff
@@ -396,6 +396,9 @@ from a source to a destination and then deleting the directory from the source.
 The entire directory's contents, including files and subdirectories, will be
 moved to the destination and then removed from the source.
 
+View this flow in the `Globus web app. <https://app.globus.org/flows/9123c20b-61e0-46e8-9469-92c999b6b8f2/definition>`_
+(You may need to log in first.)
```
jakeglobus, line 400: I don't know if this matters, but this may also require providing consent.

jakeglobus (approved):
I verified the links are correct

docs/source/authoring_flows.rst
rudyardrichter: yeah, that whole procedure is what was meant here, I think you do need to consent if you haven't used flows before. do you think there's a better wording?

jakeglobus: I wouldn't worry about it too much. Looking at it now, I think the wording is fine.

PR #72, opened 2021-06-04
Set tokenfile to R/W by owner only

jimPruyne:
Inspired by announcements related to Search client which may also
impact other uses of the globus sdk tokenstore (which maybe should
include this client at some point as well).

### DIFF
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -156,6 +156,10 @@ class TokenCache:
 
         if self.modified:
             with open(self.token_store, "w") as f:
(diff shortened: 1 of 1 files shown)


urimandujano (commented):
If I understand correctly - this will update any existing JSON storage file permissions the next time the token storage gets saved?

jimPruyne:
> If I understand correctly - this will update any existing JSON storage file permissions the next time the token storage gets saved?

That is correct. I think that's the best we can do at this point and it doesn't seem too bad.

urimandujano (approved):

PR #80, opened 2021-07-26
Updates to match Flows service updates

jimPruyne:
Some of these include:

- Renaming of role values and migration of filtering to filter_role
instead of filter_roles (plural). Impacts both SDK and CLI.

- Update documentation to talk about different roles, and also
alternative methods of specifying values in expressions and the new
method of catching timeouts.

- Where identities are entered for roles, allow the Globus Auth
username to be used in place of the full urn with uuid. Thus,
john_doe@state.edu could be used to specify a user with a role rather
than using the UUID. Docs updated to guide people as such.

- A few other cleanups here and there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,14 +42,14 @@ requirements.txt: poetry.lock
 
 # linting is flake8
 lint: develop
-	$(POETRY) run isort --check $(LINT_PATHS)
+	$(POETRY) run isort --check .
 	$(POETRY) run black --check $(LINT_PATHS)
 	$(POETRY) run flake8
 	$(POETRY) run mypy
 
 # formatting is black
 autoformat: develop
-	$(POETRY) run isort $(LINT_PATHS)
+	$(POETRY) run isort .
 	$(POETRY) run black $(LINT_PATHS)
 
 # typecheck with mypy
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,13 @@ VIRTUAL_ENV = .venv
 PYTHON_VERSION ?= python3.6
 POETRY ?= poetry
 
-.PHONY: lock test build clean help lint develop format typecheck lint_all api-docs docs
+.PHONY: all lock test build clean help lint develop format typecheck lint_all api-docs docs
+LINT_PATHS = globus_automate_client/
 
 # settings from .pytest.cfg file
 PYTEST_OPTS?=-c .pytest.cfg
+
+all: autoformat lint test docs
 
 help:
 	@echo "These are our make targets and what they do."
@@ -39,11 +42,15 @@ requirements.txt: poetry.lock
 
 # linting is flake8
 lint: develop
-	$(POETRY) run flake8 globus_automate_client
+	$(POETRY) run isort --check $(LINT_PATHS)
+	$(POETRY) run black --check $(LINT_PATHS)
+	$(POETRY) run flake8
+	$(POETRY) run mypy
 
 # formatting is black
-format: develop
-	$(POETRY) run black -q globus_automate_client tests
+autoformat: develop
+	$(POETRY) run isort $(LINT_PATHS)
+	$(POETRY) run black $(LINT_PATHS)
 
 # typecheck with mypy
 typecheck: develop
@@ -71,7 +78,7 @@ clean:
 	rm -rf tar-source
 	rm -rf dist
 
-docs:
+docs: develop
 	poetry run typer globus_automate_client/cli/main.py utils docs --name "globus-automate" --output cli_docs.md;
 	pandoc --from markdown --to rst -o docs/source/cli_docs.rst cli_docs.md;
 	rm cli_docs.md;
--- a/docs/source/authoring_flows.rst
+++ b/docs/source/authoring_flows.rst
@@ -60,7 +60,6 @@ detail below:
     {
       "Type": "Action",
       "ActionUrl": "<URL to the Action, as defined above for various Actions>",
-      "ActionScope": "<Scope String for the Action, as defined above for various Actions>",
       "WaitTime": 3600,
       "ExceptionOnActionFailure": true,
       "RunAs": "User",
@@ -88,6 +87,7 @@ detail below:
         }
       ],
       "Next": "FollowingState",
+      "ActionScope": "<Scope String for the Action, as defined above for various Actions>",
       "End": true
     }
 
@@ -98,8 +98,6 @@ in further sections below this enumeration.
 * ``Type`` (required): As with other States defined by the States Language, the ``Type`` indicates the type of this state. The value ``Action`` indicates that this state represents an Action invocation.
 
 *  ``ActionUrl`` (required): The base URL of the Action. As defined by the Action Interface, this URL has methods such as ``/run``, ``/status``, ``/cancel`` and so on defined to manage the life-cycle of an Action. The Action Flow state manages the life-cycle of the invoked Action using these methods and assumes that the specific operations are appended to the base URL defined in this property. For Globus operated actions, the base URLs are as defined previously in this document.
-
-*  ``ActionScope`` (optional): The scope string to be used when authenticating to access the Action. Users of the Flow in which this definition occurs will be required to consent to the Flow use of this scope on their behalf. This may be omitted when the Action, including another Flow, has a ``visible_to`` value including ``public`` in which case the Flows service will be able to introspect the Action via the ``ActionUrl`` and fill in the scope automatically. Any value provided here will override any use of introspection.
 
 *  ``WaitTime`` (optional, default value ``300``): The maximum amount time to wait for the Action to complete in seconds. Upon execution, the Flow will monitor the execution of the Action for the specified amount of time, and if it does not complete by this time it will abort the Action. See `Action Execution Monitoring`_ for additional information on this. The default value is ``300`` or Five Minutes.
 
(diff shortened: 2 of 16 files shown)


globus_automate_client/helpers.py
```diff
@@ -0,0 +1,33 @@
+import typing as t
+
+
+def merge_lists(*args) -> t.Optional[t.List]:
```
urimandujano, line 4: This func is a little confusing and I think that's because it's being flexible to multiple inputs. What is it being specifically used for? I'm wondering if we can pin down exactly what we need it to handle so that we can make its args + implementation more rigid.

globus_automate_client/flows_client.py
```diff
@@ -675,73 +753,91 @@ def enumerate_actions(
...
+            the Run (equivalent to ``enumerate_runs``).
 
         :param flow_scope: The scope associated with the Flow ``flow_id``. If
             not provided, the SDK will attempt to perform an introspection on
             the Flow to determine its scope automatically
 
-        :param statuses: A list of statuses used to filter the Actions that are
-            returned by the listing. Returned Actions are guaranteed to have one
-            of the specified ``statuses``. Valid values are:
+        :param  statuses: The same as in ``enumerate_runs``.
```
urimandujano, line 798: Is there a way to make the `enumerate_runs`text show up as a link to the relevant docs? That should make it easier for someone trying to get a handle on this signature.

globus_automate_client/flows_client.py
```diff
@@ -205,14 +229,14 @@ def deploy_flow(
...
-        administered_by: List[str] = [],
+        keywords: Iterable[str] = (),
+        flow_viewers: Iterable[str] = (),
+        flow_starters: Iterable[str] = (),
+        flow_administrators: Iterable[str] = (),
         subscription_id: Optional[str] = None,
         input_schema: Optional[Mapping[str, Any]] = None,
         validate_definition: bool = True,
-        validate_input_schema: bool = True,
+        validate_schema: bool = True,
```
urimandujano, line 239: I think we had this JSONSchema validation opt-out-able because we were trying to get a feel for how it was working. It's been a few months and no ones complained. I'm open to removing these flags altogether and making validation an SDK staple (:

globus_automate_client/cli/flows.py
```diff
@@ -625,11 +756,12 @@ def flow_run(
 
 
 @app.command("action-list")
+@app.command("run-list")
 def flow_actions_list(
     flow_id: str = typer.Option(
-        ...,
-        help="The ID for the Flow which triggered the Action.",
-        prompt=True,
+        None,
```
urimandujano, line 762:
```
flow_id: Optional[str] = ...
```

globus_automate_client/cli/flows.py
```diff
@@ -327,9 +432,9 @@ def flow_update(
             subtitle,
             description,
             keywords,
-            visible_to,
-            runnable_by,
-            administered_by,
+            flow_viewer,
```
urimandujano, line 435: It looks like we end up not using any of the aliases in the actual SDK call. Should we incorporate the aliases somehow? I'm ok if we decide to drop the aliases in this release.

globus_automate_client/cli/flows.py
```diff
@@ -132,33 +181,65 @@ def flow_deploy(
...
         None,
-        help="A principal which may view this Flow. The principal value is the user's "
-        "or group's UUID prefixed with either 'urn:globus:groups:id:' or "
-        "'urn:globus:auth:identity:'. The special value of 'public' may be used to "
-        "indicate that any user can view this Flow. [repeatable]",
         callback=principal_or_public_validator,
+        hidden=True,
     ),
-    administered_by: List[str] = typer.Option(
+    flow_starter: List[str] = typer.Option(
```
urimandujano, line 206: I don't think we need to worry about so many aliases on the CLI. I would be ok just keeping the starter, administrator, owner etc. CLI options for the sake of keeping the interface simpler, esp on this minor release update.

globus_automate_client/cli/callbacks.py
```diff
@@ -37,31 +46,32 @@ def _base_principal_validator(
...
-    return principals
+        if special_vals and p in special_vals or re.match(_principal_urn_regex, p):
+            valid_principals.append(p)
+        else:  # Try to do a lookup of the identity
+            if auth_client is None:
+                auth = get_authorizer_for_scope(
+                    "urn:globus:auth:scope:auth.globus.org:view_identities"
+                )
+                auth_client = AuthClient(authorizer=auth)
+            auth_resp = auth_client.get_identities(usernames=p)
```
urimandujano, line 65: Will this raise a GlobusAPIError if p isn't found? If so I think we can catch that here to keep this from crashing.

As a side note, will this end up getting all linked identities for a principal `p`? If so I'm wary of doing that look up. I can see a user specifying an identity on --manage-by, and in this implementation all of the user's linked identities would be passed onto the manage_by field of an Action/Run, which feels like a client side version of the the work we did server side to prevent linked identities being added by default.

docs/source/index.rst
```diff
@@ -1,14 +1,10 @@
 Globus Automate - CLI and Python SDK
 ====================================
 
-This is an experimental, unsupported command line interface and Python package
+This is the command line interface and Python package
```
urimandujano, line 4: 👀

docs/source/authoring_flows.rst
```diff
@@ -213,30 +219,48 @@ The syntax of an expression paramter takes the following form:
...
+        "sub_val1": "embedded",
+        "sub_val2": "also_embedded"
+      }
+    }
+
+The ``state_val`` values can be specified in one of two ways with the first syntax being preferred for long term support in the service.
+
+1. Simple names of the properties in the state of the running flow allowing for indexing into lists and into embedded objects similar to Python. Thus, the following would be a valid expression: ``foo + ' ' + object_val.sub_val1`` which would yield the string ``bar embedded``. Note the use of ``+`` to mean string concatenation and the dot-separated naming of the field of the object.
+
+2. The `JsonPath <https://goessner.net/articles/JsonPath/>`_ syntax
```
urimandujano, line 249: I wonder if we should still advertise this as an option. I guess it's good for folks who are already working with Flows that use the JSONPath syntax.

docs/source/authoring_flows.rst
```diff
@@ -140,7 +140,13 @@ both cases.
 
 For ``Parameters``, a list with special property name ``__Private_Parameters``
 may be placed in the ``Parameters`` object indicating which other Parameters
-should be protected. For simplicity, the values in the ``__Private_Parameters``
+should be protected. These values will be protected in two ways:
+
+* Users that lookup the Flow in the service will not see the ``Parameters`` which are specified in the ``__Private_Parameters`` list unless the have the ``flow_administrator`` or ``flow_owner`` role on the Flow.
```
urimandujano, line 145:
```suggestion
* Users that lookup the Flow in the service will not see the ``Parameters`` which are specified in the ``__Private_Parameters`` list unless they have the ``flow_administrator`` or ``flow_owner`` role on the Flow.
```

docs/source/authoring_flows.rst
```diff
@@ -123,6 +121,8 @@ in further sections below this enumeration.
 
 *   ``Catch``: When Actions end abnormally, an Exception is raised. A ``Catch`` property defines how the Exception should be handled by identifying the Exception name in the ``ErrorEquals`` property and identifying a ``Next`` state to transition to when the Exception occurs. If no ``Catch`` can handle an exception, the Flow execution will abort on the Exception. A variety of exception types are defined and are enumerated in `Managing Exceptions`_.
 
+*  ``ActionScope`` (optional): The scope string to be used when authenticating to access the Action. In most cases, this values is unneeded because the required scope can be determined by querying the Action Provider using the provided ``ActionUrl``. If you are using a non-standard compliant Action which does not publish its ``scope``, this can be provided to avoid attempting to query the non-compliant Action provider.
+
```
urimandujano, line 125: I like this update to the description

Makefile
```diff
@@ -39,11 +42,15 @@ requirements.txt: poetry.lock
...
+	$(POETRY) run black --check $(LINT_PATHS)
+	$(POETRY) run flake8
+	$(POETRY) run mypy
 
 # formatting is black
-format: develop
-	$(POETRY) run black -q globus_automate_client tests
+autoformat: develop
+	$(POETRY) run isort $(LINT_PATHS)
+	$(POETRY) run black $(LINT_PATHS)
```
urimandujano, line 53: Some comment here with removing the LINT_PATHS in favor of pyproject

Makefile
```diff
@@ -39,11 +42,15 @@ requirements.txt: poetry.lock
 
 # linting is flake8
 lint: develop
-	$(POETRY) run flake8 globus_automate_client
+	$(POETRY) run isort --check $(LINT_PATHS)
+	$(POETRY) run black --check $(LINT_PATHS)
```
urimandujano, line 46: We're probably better off defining the lint paths in pyproject.toml and letting the tools read their own configuration when we run them. It at least removes the need for the extra variable here and keeps the project config slightly more centralized.

globus_automate_client/helpers.py
jimPruyne: I kinda agree this is a bit wonky. I set up the input signature for this for a very particular use case of wanting to merge lists where some were lists as params to a function and some were potentially in kwargs to that function. So, I wanted to just be able to pass the list of values concisely in the caller and have this merge those lists.

This was all driven by trying to use kwargs for backward compatibility where we were changing the name of some parameters.

So, bottom-line, I agree this is on the spectrum of wonky to "very purpose specific". We can further discuss towards something more rigid/better defined if you'd like.

globus_automate_client/flows_client.py
jimPruyne: That's a good suggestion. I don't now know of any good sphinx-y trick to do that. Rudy's comment the other day about writing something to pull the docstring from one function and use it to create the docstring of another might be a different solution. We probably end up with more verbose docs as there'd be replication, but maybe that's better than an open-ended sort of reference?

Again, I'm open to suggestions on how to improve something I know is non-ideal.

globus_automate_client/flows_client.py
jimPruyne: I think I'm good with continuing to make opt-in for validation the default, but allow people to override if they really want to. Can I imagine a situation where validation would blow up on the client but then succeed on the server? Especially in the case of the schema I cannot since I believe we're using the exact same validation from `jsonschema`, but I guess it feels like one of those things where I fear someone will complain that the client is limiting them somehow. I dunno...

globus_automate_client/cli/callbacks.py
jimPruyne: Turns out this doesn't throw an error if the identity doesn't exist, so we should make sure it does do that.

jimPruyne: fixed!!

urimandujano (approved):

urimandujano:
@jimPruyne I'm happy w/ this if you are!

PR #84, opened 2021-08-24
Unifies CLI output display and adds tabular display

urimandujano:
This PR adds support for displaying list data as tabular output [see screenshots below]. Notice that the tabular outputs play well with the other options available on the CLI such as filtering, verbose output, and watching for updates. Only a subset of the API's returned data gets formatted as a table due to the limits of a terminal window but the fields we display can be tweaked (or processed as in the case of the datetime fields in the screenshots). For the full data to be parsed, the output needs to be formatted as either JSON or YAML. On all commands, if the output is specified as JSON, external tools (like JQ) can parse the output and pull specific fields if needed.

The general approach was to separate the `Result` and `Renderer` into distinct objects, and then manage both via a `RequestRunner`. The runner will execute a given callable and then render the result according to the various options (increased verbosity, willingness to poll, output format). A `CompletionDetetector` can be passed as an argument to the `RequestRunner` which is used to determine when a `Result` can be considered complete in the event that polling occurs. Endpoints that have produce list data (such as Flow-list) can also pass a `DisplayFields` object that are used to determine which fields are displayed for a given list item and how to display them.

This PR adds some auto parsing of yaml/json files, no longer requiring users to indicate what type of input they're providing. It also adds an example flow for creating a Run that will become INACTIVE due to ConsentsRequired. It also updates the Queues client to match changes made to the Queue update API endpoint.

Run log output:
Run-Log-Output

Flow list output [with filtering]:
Flow-List-Output

Run enumerate output [with options]:
Run-Enumerate-Output

Flow action-list output [with authentication]:
Flow-Action-List-Output

### DIFF
--- a/globus_automate_client/cli/constants.py
+++ b/globus_automate_client/cli/constants.py
@@ -1,9 +1,7 @@
 import enum
-import functools
-import json
+from itertools import chain
 
 import typer
-import yaml
 from globus_sdk import GlobusHTTPResponse
 
 from globus_automate_client.graphviz_rendering import (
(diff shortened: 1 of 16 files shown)


braumann:
Nice!  Could user uuid be easily replace with user's auth identity (e.g. replace uuid with braumann@globus.org)?

globus_automate_client/cli/rich_helpers.py
```diff
@@ -0,0 +1,435 @@
...
+    prehook: Optional[Callable[[List[Dict[str, Any]]], List[Dict[str, Any]]]] = None
+
+
+class RunListDisplayFields(DisplayFields):
+    """
+    This object defines the fields and display style of a Run listing into a
+    table.
+    """
+
+    fields = [
```
jimPruyne, line 116: If we decided to make this dynamic later where a user could do something like `--fields action_id, start_time, end_time` or what have you to specify which fields should be in the tabular output, we could probably do that here rather than this static list.

That'd be a follow-on feature, but if you see anything in the existing infrastructure that would make it easier to parameterize later, perhaps we try to drop it in right now.

globus_automate_client/cli/rich_helpers.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+class FlowListDisplayFields(DisplayFields):
+    """
+    This object defines the fields and display style of a Flow listing into a
+    table.
+    """
+
+    fields = [
+        Field("title", ""),
```
jimPruyne, line 144: adding another idea when I see this again: Could the `Field` be an Enum of some sort where these name to transformation things are properties of the field in the Enum. Thus, maybe, that Enum could also end up being the type for the Typer parameter for `--output-fields` or some such.

Again, I don't think we have to roll that into this PR, but mgiht be something to address soon after?

globus_automate_client/cli/rich_helpers.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+
+    def __init__(
+        self,
+        response: Union[GlobusHTTPResponse, GlobusAPIError, str],
+        detetector: Optional[Type[CompletionDetetector]] = None,
+    ):
+        self.result = response
+        if detetector is None:
+            self.detetector: Type[CompletionDetetector] = ActionCompletionDetector
```
jimPruyne, line 229: Could this just be the default on the method rather than making it `Optional` and then testing for `None`?

globus_automate_client/cli/rich_helpers.py
```diff
@@ -0,0 +1,435 @@
...
+        else:
+            self.detetector = detetector
+
+        self.is_api_error = isinstance(response, GlobusAPIError)
+        self.data: Dict[str, Any]
+        if isinstance(response, str):
+            self.data = {"result": response}
+        elif isinstance(response, GlobusAPIError):
```
jimPruyne, line 237: could re-use `self.is_api_error` here?

jimPruyne (commented):
I think most of my comments are cosmetic. There's a lot of nice cleanup here and most of my comments are perhaps chances for additional cleanup here and there. Perhaps you've already considered many of them.

I guess the real meat of this, where the table is formatted to strings eludes me. I'm guessing a `Typer` or other helper library is doing the heavy lifting, and perhaps makes it so simple that I miss it.

Anyway, the examples look great, and we already have some positive feedback from the RC, so I consider my comments as optional and we can decide which to address now/ignore/address in a follow-up.

globus_automate_client/cli/constants.py
```diff
@@ -12,41 +10,126 @@
...
-class FlowDisplayFormat(str, enum.Enum):
     json = "json"
-    graphviz = "graphviz"
+    yaml = "yaml"
+    table = "table"
     image = "image"
+    graphiz = "graphiz"
+
+    def visualize(self, flow_log: GlobusHTTPResponse, flow_def: GlobusHTTPResponse):
+        if self == "image":
```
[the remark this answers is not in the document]
urimandujano: `is` always gets me thinking of identity equality which seems shaky since these are strings. I feel like == is clearer in this case.

globus_automate_client/cli/flows.py
```diff
@@ -247,13 +167,13 @@ def flow_deploy(
...
     ),
-    flows_endpoint: str = _flows_env_var_option,
+    flows_endpoint: str = flows_env_var_option,
     verbose: bool = verbosity_option,
-    input_format: InputFormat = typer.Option(
-        InputFormat.json,
-        "--input",
-        "-i",
-        help="Input format.",
+    output_format: OutputFormat = typer.Option(
```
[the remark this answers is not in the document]
urimandujano: Sure, I did like the `flows_env_var_option` refactor.

globus_automate_client/cli/rich_helpers.py
urimandujano: We could probably define (field:default:transformation) tuples for each field we want to support and let the user decide which combination of fields they want to display. Right now I don't think we need that complexity/flexibility tradeoff.

globus_automate_client/cli/rich_helpers.py
urimandujano: Could be! I think it depends on what users want to do, ideally we'd be able to have a single generic view that works and makes everyone happy. This CLI bit of code is already more complicated than I'd originally hoped.

globus_automate_client/cli/rich_helpers.py
urimandujano: Good suggestion for simplifying the interface. Made the change + propagated it out.

urimandujano:
@jimPruyne I finally got around to making updates from the review. Let me know how you're feeling about these changes and I can go ahead and push the green button

jimPruyne (approved):

jimPruyne:
> @jimPruyne I finally got around to making updates from the review. Let me know how you're feeling about these changes and I can go ahead and push the green button

Looks good! Let's schedule a 0.13 (?) release within the next day or so.

PR #87, opened 2021-09-20
Resize flow list table to insure uuid is complete

jimPruyne:
Without the whole flow uuid being displayed, it is hard to make use of the output from the flow list operation. This tweaks some field widths and, unfortunately, removes some fields from the output to make sure the complete uuid is shown.

### DIFF
--- a/globus_automate_client/cli/rich_helpers.py
+++ b/globus_automate_client/cli/rich_helpers.py
@@ -21,6 +21,9 @@ from .constants import OutputFormat
 from .helpers import get_http_details
 from .rich_rendering import cli_content
 
+_title_max_width = 25
+_uuid_min_width = 36
+
 
 def humanize_datetimes(dt: str) -> str:
     return arrow.get(dt).humanize()
@@ -81,6 +84,8 @@ class Field:
         name: str,
         default: str,
         transformation: Optional[Callable[[str], str]] = None,
+        min_width: Optional[int] = None,
+        max_width: Optional[int] = None,
     ) -> None:
         """
         name: The name of the field which contains the data for display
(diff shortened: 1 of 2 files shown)


urimandujano (changes requested):
I think a nicer option would be to use the `overflow="fold"` option on `add_column`. That should change a too-long-to-display UUID from truncating with ellipses to wrapping onto the next line, without dropping any fields.

jimPruyne:
> I think a nicer option would be to use the `overflow="fold"` option on `add_column`. That should change a too-long-to-display UUID from truncating with ellipses to wrapping onto the next line, without dropping any fields.

Could you try that out? I also note that the fields being displayed now are essentially the same as the main globus cli provides in output for their endpoint search, so having similar sets of fields seems like good consistency particuarly as we head toward merging with that.

jimPruyne:
> I think a nicer option would be to use the overflow="fold" option on add_column. That should change a too-long-to-display UUID from truncating with ellipses to wrapping onto the next line, without dropping any fields.

@urimandujano I thought about this again later, and realized this is perhaps not a good idea, and that that `min_width` is the way to go. My reasoning is that a common use for the output is to then do a cut and paste from the terminal window for use in further commands. Even if it is wrapped to another line, it makes doing all that in a single cut-and-paste impossible (I think?). So, I think allocating more screen width to the uuid, at the expense of other columns, is a worthwhile trade-off. Thus, I think, this is probably worth reviewing as is.

urimandujano:
I'm torn on this. I just took a look at the output using the fixed width/ removing the other two columns and I don't think it gives a valuable enough view of the Flows listing outputs. Flow title, flow id and flow owner just don't seem to offer enough info by themselves. Asking users to resize their windows to get a continuous Flow ID / do a two part copy paste doesn't seem bad enough to cut down on the fields returned.

PR #90, opened 2021-10-08
Make watches tabular by default and fix some bugs

kurtmckee:
Fix several bugs (a crash, an infinite loop, an incorrect value constraint) and make tabular output the default when watching flow output.

### DIFF
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -585,7 +585,18 @@ def flow_run(
     ),
     flows_endpoint: str = flows_env_var_option,
     verbose: bool = verbosity_option,
-    output_format: OutputFormat = output_format_option,
+    output_format: ListingOutputFormat = typer.Option(
+        None,
+        "--format",
+        "-f",
+        help=(
+            "Output display format."
+            " If --watch is enabled then the default is 'table',"
+            " otherwise 'json' is the default."
+        ),
+        case_sensitive=False,
+        show_default=False,
+    ),
     label: str = typer.Option(
         ...,
         "--label",
@@ -596,7 +607,10 @@ def flow_run(
         False,
         "--watch",
         "-w",
-        help="Continuously poll this Action until it reaches a completed state.",
+        help=(
+            "Continuously poll this Action until it reaches a completed state."
+            " If enabled the default output format is 'table'."
+        ),
         show_default=True,
     ),
     dry_run: bool = typer.Option(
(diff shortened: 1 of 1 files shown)


globus_automate_client/cli/flows.py
```diff
@@ -585,7 +585,18 @@ def flow_run(
     ),
     flows_endpoint: str = flows_env_var_option,
     verbose: bool = verbosity_option,
-    output_format: OutputFormat = output_format_option,
+    output_format: ListingOutputFormat = typer.Option(
+        None,
```
urimandujano, line 589: The first value to typer.Option is what gets used as the default if the command line flag isn't explicitly set. What do you think of setting that to `ListingOutputFormat.json` rather than `None` and updating the `output_format` logic below?

globus_automate_client/cli/flows.py
```diff
@@ -940,15 +966,15 @@ def flow_action_log(
         callback=url_validator_callback,
     ),
     reverse: bool = typer.Option(
-        # Defaulting to any boolean value will reverse output - so we use None
-        None,
+        False,
```
urimandujano, line 969: Based on how the API interprets any truthy value as a request to reverse the order, maybe we should just update the type annotation here that it can be a `None` value. Or we could implement the correct behavior here and try to update the API to behave correctly. That could be/should be a good story!

globus_automate_client/cli/flows.py
kurtmckee: If I set a default value then I think it will be impossible to distinguish between the default value and a user's explicit override. Doing it this way allows the user to explicitly watch with JSON or YAML output, which maintains existing functionality:

```
# With a default of None
# ----------------------

globus-automate flow run                # output_format=None so we can pick a default
globus-automate flow run --format json  # output_format="json" so the user explicitly wants JSON


# With a default of "json"
# ------------------------

globus-automate flow run                # output_format="json"
globus-automate flow run --format json  # output_format="json"
```

I don't know if it's truly useful to keep that functionality, though. Should we forbid watching in JSON and YAML formats?

globus_automate_client/cli/flows.py
kurtmckee: I think I addressed this issue at its root failure point, in `globus_automate_client.action_client.ActionClient.log()`.

It should now be possible to use `True` and `False` throughout the code without special knowledge that only `None` actually works.

globus_automate_client/cli/flows.py
kurtmckee: Thanks Uriel! 😄

PR #92, opened 2021-10-20
Use single sources of truth

kurtmckee:
This patch adjusts these items:

* Remove trove classifiers that Poetry will automatically add based on package metadata.
* Bump `typer-cli` to 0.0.12, which removes its dependency on an old version of `importlib-metadata`.
* Add `importlib-metadata` as a dependency for Python 3.6 and 3.7.
* Use the package metadata as the single source of version truth.


kurtmckee:
As this affects trove classifiers and dependencies I'd like to have one additional approval. @rudyardrichter or @jimPruyne?

PR #95, opened 2021-10-29
Adds docs on new APs and expression functions

jimPruyne:
We add docs related to:

- The Collection info AP

- The make directory AP

- The updates to the ls AP

- That there are functions which can be invoked within expressions,
and the following functions:

  - `getattr`
  - `is_present`
  - `pathsplit`


urimandujano:
This looks good. I noticed that it's not ready for review -- were you meaning to come back to this? By the way, can you try out the new scriv changelog process? This would benefit from getting tracked for release notes.

jimPruyne:
@urimandujano if the base docs look ok to you, we can take this through the review process. I've added the changelog file via scriv.

urimandujano (approved):
This looks good thanks!

PR #96, opened 2021-10-29
Add workflow to check for changelog fragments

rudyardrichter:
Per our discussion a while ago, this is a simple workflow to check for the addition of changelog fragments in a PR, or else a PR label (which I added for this one).

Here we're checking for `*.rst` files specifically, but honestly it could just be any new files in `changelog.d` which might make it easier to duplicate for other repositories, though easier to miss if someone accidentally creates a fragment in the wrong format.

### DIFF
--- a/.github/workflows/has_changelog.yaml
+++ b/.github/workflows/has_changelog.yaml
@@ -1,0 +1,22 @@
+name: has_changelog
+on:
+  - pull_request
+
+jobs:
+  check_has_news_in_changelog_dir:
+    if: ${{ ! contains(github.event.pull_request.labels.*.name, 'no-news-is-good-news') }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:  # do a deep fetch to allow merge-base and diff
+          fetch-depth: 0
+      - name: check PR adds a news file
+        run: |
+          news_files="$(git diff --name-only "$(git merge-base origin/master "$GITHUB_SHA")" "$GITHUB_SHA" -- changelog.d/*.rst)"
+          if [ -n "$news_files" ]; then
+            echo "Saw new files. changelog.d:"
+            echo "$news_files"
+          else
+            echo "No news files seen"
+            exit 1
+          fi


.github/workflows/has_changelog.yaml
```diff
@@ -0,0 +1,22 @@
...
+  check_has_news_in_changelog_dir:
+    if: ${{ ! contains(github.event.pull_request.labels.*.name, 'no-news-is-good-news') }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:  # do a deep fetch to allow merge-base and diff
+          fetch-depth: 0
+      - name: check PR adds a news file
+        run: |
+          news_files="$(git diff --name-only "$(git merge-base origin/main "$GITHUB_SHA")" "$GITHUB_SHA" -- changelog.d/*.rst)"
```
kurtmckee, line 15: In the past I've used a command like this to ask git for a list of new files introduced between two branches:

```shell
git diff --name-only --diff-filter=A branch1 branch2
```

I think you're right: it's likely that the only requirement is that *any* new file exists, so this command may be easier to explain. What are your thoughts?

rudyardrichter: There's a surprising number of things to unpack there, now that you point this out. I'll add comments for any you think are useful.
- I think we don't necessarily want `--diff-filter` in this case, as a PR which removes chunks or renaming files still merits changelog entries.
- We can't necessarily just diff this branch and origin/master because the author may have started their branch/PR from an older commit than latest origin/master, so that is what `merge-base` accomplishes. `merge-base` should effectively be fixed over all commits on the same pull request (save for rebasing etc.), similarly to a branch name, and the `$GITHUB_SHA` variable here (as I understand it) is for a "phantom merge" for what would happen if one were to merge the PR right then.

rudyardrichter: I noticed while writing the above that the workflow is using `main` instead of `master`, so that's now fixed.

kurtmckee: Whoops, I meant to point that sample command at the `changelog.d/` directory, so only new fragments introduced in the branch would be displayed, like

```
git diff --name-only --diff-filter=A main "$GITHUB_SHA" -- changelog.d/
```

I think that if origin/main has received additional merges that wouldn't trigger any issues: the diff-filter is limiting the results to files added in the current branch compared to origin/main. Files introduced by other recently-merged branches wouldn't show up in the results.

I had no idea about the `$GITHUB_SHA` phantom merge situation, though!

kurtmckee: Example script to confirm behavior:

```shell
# Initial setup.
git init diff-filter-test
cd diff-filter-test
mkdir changelog.d
echo "a" > changelog.d/existing-fragment.txt
git add changelog.d/existing-fragment.txt
git commit -m "initial commit"

# Create a new branch.
git checkout -b feature-branch
echo "b" > changelog.d/my-new-fragment.txt
git add changelog.d/my-new-fragment.txt
git commit -m "new feature"

# Create an additional commit on main.
git checkout main
echo "c" > changelog.d/your-new-fragment.txt
git add changelog.d/your-new-fragment.txt
git commit -m "unrelated changes"

# Check the output.
echo ""
echo "Result:"
echo "-------"
echo ""
git diff --name-only --diff-filter=A main feature-branch -- changelog.d/
```

rudyardrichter: I'm imagining the following case, earliest first chronologically:
- (1) Branch/fork starts here
- (2) Commits merged to main delete some files (like changelog fragments!)
- (3) Latest origin/main

A diff from (3) to (1), including `--diff-filter=A`, reports the deleted files as added.

It's slightly pathological, because you should rebase the branch on latest main, but un-pathological because deleting changelog fragments like that is part of the process with scriv. I _think_ that this `merge-base` method is the best way to accomplish "show me the changes this PR means to introduce".

kurtmckee: That clears it up for me. Thanks for the explanation!

kurtmckee (approved):

urimandujano:
Thanks for the explanation @rudyardrichter, that's a hefty git line with a new git command for me so having it explained was helpful. If this works out well I'd be happy with pulling this workflow from our other repos and have a sort of standardized changelog process.

rudyardrichter:
A note for posterity: if we want to reuse this elsewhere then we may have to figure out how to parameterize `main` vs. `master`.

PR #97, opened 2021-10-29
Upgrade to SDK v3 (original / split into smaller PR's)

kurtmckee:
This upgrades the automate client to SDK v3. It's possible that this is best reviewed by clicking on the first commit and working through the changes commit by commit.

I don't think that this work is complete, and certainly not as tested as it needs to be, but I'd like to get feedback on the work so far. From the changelog fragment:

Features
--------

-   Upgrade to Globus SDK v3.
-   Add code linting, documentation build testing, and a bunch of unit tests.
    (Code coverage of modules currently imported during testing is 73%.)

Bugfixes
--------

-   Fix a bug that prevented the Flows client from properly validating flow definition states in lists.
-   Fix a bug that could allow the Flows authorizer to be lost if an exception was raised.
    (Authorizer swaps are now handled using a context manager.)
-   Prevent empty values from appearing in query parameters or JSON submissions.
-   Fix a bug that prevented the input schema of an existing Flow from being set to an all-permissive JSON schema.

### DIFF
--- a/changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
+++ b/changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
@@ -1,0 +1,19 @@
+Features
+--------
+
+-   Upgrade to Globus SDK v3.
+
+Bugfixes
+--------
+
+-   Fix a bug that prevented the Flows client from properly validating flow definition states in lists.
+-   Fix a bug that could allow the Flows authorizer to be lost if an exception was raised.
+    (Authorizer swaps are now handled using a context manager.)
+-   Prevent empty values from appearing in query parameters or JSON submissions.
+-   Fix a bug that prevented the input schema of an existing Flow from being set to an all-permissive JSON schema.
+
+Other
+-----
+
+-   Add code linting, documentation build testing, and a bunch of unit tests.
+    (Code coverage of modules currently imported during testing is 73%.)
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -72,7 +72,7 @@ class TokenCache:
         self.modified = False
 
     @property
-    def tokens_for_environment(self) -> Dict[str, TokenSet]:
+    def tokens_for_environment(self):
         """
         We will sub-key the full token set for environments other than production
         """
@@ -142,7 +142,7 @@ class TokenCache:
             )
 
     @staticmethod
-    def _make_jsonable(tokens: TokensInTokenCache) -> Dict[str, Any]:
+    def _make_jsonable(tokens) -> Dict[str, Any]:
         serialized: Dict[str, Any] = {}
         for k, v in tokens.items():
             if isinstance(v, TokenSet):
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -9,14 +9,18 @@ from typing import Any, Callable, Dict, List, NamedTuple, Optional, Set, Union
 
 import click
 import typer
-from globus_sdk import AuthClient, GlobusAPIError, NativeAppAuthClient
-from globus_sdk import OAuthTokenResponse
+from globus_sdk import (
+    AuthAPIError,
+    AuthClient,
+    GlobusAPIError,
+    NativeAppAuthClient,
+    OAuthTokenResponse,
+)
 from globus_sdk.authorizers import (
     AccessTokenAuthorizer,
     GlobusAuthorizer,
     RefreshTokenAuthorizer,
 )
-from globus_sdk import AuthAPIError
 
 from globus_automate_client.action_client import ActionClient
 
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -289,7 +289,7 @@ def get_authorizers_for_scopes(
                     grant_response: OAuthTokenResponse, *args, **kwargs
                 ):
                     token_cache.update_from_oauth_token_response(
-                        grant_response, set([scope])
+                        grant_response, {scope}
                     )
 
                 authorizer = RefreshTokenAuthorizer(
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -10,13 +10,13 @@ from typing import Any, Callable, Dict, List, NamedTuple, Optional, Set, Union
 import click
 import typer
 from globus_sdk import AuthClient, GlobusAPIError, NativeAppAuthClient
-from globus_sdk.auth.token_response import OAuthTokenResponse
+from globus_sdk import OAuthTokenResponse
 from globus_sdk.authorizers import (
     AccessTokenAuthorizer,
     GlobusAuthorizer,
     RefreshTokenAuthorizer,
 )
-from globus_sdk.exc import AuthAPIError
+from globus_sdk import AuthAPIError
 
 from globus_automate_client.action_client import ActionClient
 
--- a/globus_automate_client/cli/auth.py
+++ b/globus_automate_client/cli/auth.py
@@ -299,8 +299,6 @@ def get_authorizers_for_scopes(
                     expires_at=token_set.expiration_time,
                     on_refresh=refresh_handler,
                 )
-                # Force check that the token is not expired
-                authorizer.check_expiration_time()
             else:
                 authorizer = AccessTokenAuthorizer(token_set.access_token)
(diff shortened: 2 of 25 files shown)


changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
```diff
@@ -0,0 +1,15 @@
+Features
+--------
+
+-   Upgrade to Globus SDK v3.
+-   Add code linting, documentation build testing, and a bunch of unit tests.
+    (Code coverage of modules currently imported during testing is 73%.)
```
rudyardrichter, line 6: I would characterize this as a different heading ("Changes" or "Other"? I don't know if these exist by default, at least in this project) since it's not user-facing

globus_automate_client/cli/auth.py
```diff
@@ -68,16 +72,16 @@ def __init__(self, token_store: Union[pathlib.Path, str]):
...
         We will sub-key the full token set for environments other than production
         """
         environ = os.environ.get("GLOBUS_SDK_ENVIRONMENT")
-        if environ in {None, "production", "prod", "default"}:
+        if environ is None or environ in {"production", "prod", "default"}:
             return self.tokens
         environ_cache_key = TokenCache._environment_prefix + environ
         if environ_cache_key not in self.tokens:
-            self.tokens[environ_cache_key]: Dict[str, TokenSet] = {}
+            self.tokens[environ_cache_key] = {}
```
rudyardrichter, line 84: The annotation here is effectively (useful) documentation, does a linter complain about this?

globus_automate_client/queues_client.py
```diff
@@ -81,19 +84,21 @@ def update_queue(
...
+                "receivers": receivers,
+                "delivery_timeout": delivery_timeout,
+                "visibility_timeout": visibility_timeout,
+            },
+        }
         # Remove the missing values from the update operation
-        body = {k: v for k, v in body.items() if v is not None}
-        if body:
-            return self.put(f"/queues/{queue_id}", {"data": body}, **kwargs)
+        data = {k: v for k, v in data.items() if v is not None}
```
rudyardrichter, line 99: If I'm reading this correctly, this line would be a no-op since `data` only has the key-value pair `"data": {"id": ...}`?

pyproject.toml
```diff
@@ -82,3 +85,81 @@ ignore_missing_imports = true
 
 [tool.scriv]
 categories = ["Features", "Bugfixes", "Documentation"]
+
+[tool.tox]
```
rudyardrichter, line 89: Given this is a pretty large chunk maybe better to have it in a `tox.ini`? That way you'd actually get syntax highlighting/checking on this too which might be nicer.

rudyardrichter (changes requested):
I did not make a pass through the tests yet, but here are some early thoughts.

- Others may differ, but I would be in favor of squashing the commits down pretty aggressively. For instance, I might condense
```
- Resolve isort errors
- Resolve flake8 errors
- Resolve most mypy errors
```
to just one `Apply linters` or similar. Adding tests in particular I'd be in favor of squashing because it's not likely we'd want to `git bisect` those later or anything.

globus_automate_client/flows_client.py
```diff
@@ -897,16 +961,16 @@ def flow_action_update(
             argument, that gets used to run the Flow operation. Otherwise the
             authorizer_callback defined for the FlowsClient will be used.
         """
+
         payload = {}
-        if run_managers is not None:
+        if run_managers:
             payload["run_managers"] = run_managers
-        if run_monitors is not None:
+        if run_monitors:
```
rudyardrichter, line 968: I think this should be reverted. As the docstring says, an empty list actually has a particular meaning

urimandujano (commented):
This looks like a few different PRs worth of work combined into one. Can you split this PR into smaller chunks based on what the code changes are? If we go that route I think any bugfixes that get made would be good to merge and release as part of this package's final 0.12.3 release. Then we can bump our dependency on globus-sdk to 3 and call that our 0.13.0 release.

rudyardrichter (changes requested):

rudyardrichter:
I'd second Uriel, it seems like this PR has a lot going on and it would be easier to track the SDK-upgrade-specific changes if that was separate from linting changes and the like.

changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
kurtmckee: Agreed. It's not difficult to add these, it just needs to be codified in the config. For the scriv config I only copied what I saw in past release announcements:

```
categories = ["Features", "Bugfixes", "Documentation"]
```

Adding a tag like "Other" seems good to me. @jimPruyne / @urimandujano?

globus_automate_client/cli/auth.py
kurtmckee: My memory is that mypy didn't complain but PyCharm highlighted this as extraneous. It's annotating a dictionary key's value instead of the dictionary itself so it didn't seem to add additional value. Should I revert this?

globus_automate_client/queues_client.py
kurtmckee: 100% correct! I only wrote unit tests for flows and action client code so this will definitely need to be a part of the queues client tests.

pyproject.toml
kurtmckee: Good point, I'll do that!

globus_automate_client/cli/auth.py
rudyardrichter: I'd vote in favor of keeping because it made it easier to understand what `self.tokens` is doing.

globus_automate_client/flows_client.py
kurtmckee: 100% correct. I was on a kick to remove empty values. This needs to be undone.

globus_automate_client/cli/auth.py
kurtmckee: I'll get this fixed.

Separately, the token cache annotations triggered a number of errors throughout the codebase with mypy. Eventually this needs to be addressed better than what I did in 36f84f1026483cbb7b6d249b8b50fde114eec5c8.

kurtmckee:
@urimandujano and @rudyardrichter I'll work to squash the commits and stage them as separate PR's. Based on what I'm reading, the suggestion is to have separate PR's for:

* Fixing bugs (including removing empty/false-y values from query parameters and JSON data)
* Adding unit tests / linters
* Upgrading to SDK v3

They'll probably be in this order, too. I'll work to get this done.

changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
kurtmckee: I've updated the changelog fragment but it's still worthwhile to consider the categories.

kurtmckee:
I've force-pushed updates to the original branch in case any other comments are shared. I'm working on separating the test/fix/upgrade commits into separate PR's now.

rudyardrichter:
Thanks @kurtmckee, I realize it's a bunch of extra git work to get those organized, and for what will likely become the same end result, so your efforts are appreciated.

kurtmckee:
Bug fix PR is live at #98. I'd like to wait for that to get reviewed so I can build the unit testing on top of it.

kurtmckee:
It turns out the unit tests are heavily dependent on SDK v3. There is a great deal of breakage if the unit tests come next so I'm going to prep the SDK v3 changes on top of the bug fixes first. The unit tests and linting will be the final PR.

changelog.d/20211029_162341_kurtmckee_sdk_v3.rst
urimandujano: I'm not sure what to call these -- maybe "Internal", "Development" or "Meta changes"? I don't have a strong opinion but I'm willing to go with whatever feels good to you two.

urimandujano:
@kurtmckee I appreciate your extra effort on breaking these out. I took a look at #98 and that looks good to me. If we end up merging that then I think we could be good maintainers and cut a release there.

jimPruyne:
What's the current state of this PR? Is this still a to be reviewed, or has refactors / splits / etc made this obsolete, or waiting for some other step for updated review?

kurtmckee:
I left this open to maintain visibility of comments between it and the three PR's it is being split into.

kurtmckee:
This work is now fully merged through smaller commits. Closing and deleting the branch.

PR #98, opened 2021-11-02
Fix some bugs

kurtmckee:
From the changelog fragment:

Bugfixes
--------

-   Fix a bug that prevented the Flows client from properly validating flow definition states in lists.
-   Prevent empty values from appearing in query parameters or JSON submissions.
-   Fix a bug that prevented the input schema of an existing Flow from being set to an all-permissive JSON schema.
-   Fix a bug that prevented a custom authorizer from being used if attempting to list all runs of a specific flow without specifying the flow ID.
-   Fix a bug that introduced duplicate forward slashes in some API URL's.

### DIFF
--- a/globus_automate_client/flows_client.py
+++ b/globus_automate_client/flows_client.py
@@ -463,25 +463,35 @@ class FlowsClient(BaseClient):
 
         """
         self.authorizer = self.flow_management_authorizer
+
         params = {}
-        if roles is not None and len(roles) > 0:
-            params["filter_roles"] = ",".join(roles)
-        if role is not None:
+
+        # *role* takes precedence over *roles* (plural).
+        if role:
             params["filter_role"] = role
-            params.pop(
-                "filter_roles", None
-            )  # role takes precedence over roles (plural)
-        if marker is not None:
+        elif roles:
+            params["filter_roles"] = ",".join(roles)
+
+        # *marker* takes precedence over *per_page*.
+        if marker:
             params["pagination_token"] = marker
-        if per_page is not None and marker is None:
+        elif per_page:
             params["per_page"] = str(per_page)
-        if filters is not None:
-            params.update(filters)
+
         if orderings:
-            builder = []
-            for field, value in orderings.items():
-                builder.append(f"{field} {value}")
-            params["orderby"] = ",".join(builder)
+            params["orderby"] = ",".join(
+                f"{field} {value}" for field, value in orderings.items()
+            )
+
+        if filters:
+            # Prevent *filters* from overwriting reserved keys.
+            filters.pop("filter_role", None)
+            filters.pop("filter_roles", None)
+            filters.pop("orderby", None)
+            filters.pop("pagination_token", None)
+            filters.pop("per_page", None)
+            params.update(filters)
+
         return self.get("/flows", params=params, **kwargs)
 
     def delete_flow(self, flow_id: str, **kwargs) -> GlobusHTTPResponse:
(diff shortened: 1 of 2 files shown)


urimandujano (approved):
Thanks for taking the time and breaking this out - it's easier to appreciate the subtlety in these bugs and the elegance of the fixes (I'm impressed that you found so many of these!). It's unfortunate that we end up with duplicated parsing logic in some of the API calls but I think it's better not to rely on some shared behavior when the args/kwargs of a function may change.

Anyways, I think these will be very welcome corrections for anyone still relying on this package + an older version of the `globus_sdk`.

If you have the energy after breaking this out I think each bug could have its own story associated with it. This is after all something we have client's using and each fix is worth tracking.

globus_automate_client/flows_client.py
```diff
@@ -460,25 +463,35 @@ def list_flows(
...
         """
         self.authorizer = self.flow_management_authorizer
+
         params = {}
-        if roles is not None and len(roles) > 0:
-            params["filter_roles"] = ",".join(roles)
-        if role is not None:
+
+        # *role* takes precedence over *roles* (plural).
+        if role:
```
jimPruyne, line 470: The pattern for checking truthiness on one option followed by the next etc is common in this. As a rule, I am very wary of python truthiness checks. As these are parameters passed in so could have any value depending on what's coming from the caller, I'm even a bit more wary.

It _may_ be that truthiness is exactly the right check here (and elsewhere with this pattern), but I just wanted to double check this is being used with full thought to the possible values that could be passed and if, in these cases, truthiness accurately captures the intent.

jimPruyne (commented):
Overall, I think this reads well. My main comment is on the frequent use of truthiness which is something I watch out for in particular. If you feel comfortable with truthiness really capturing the condition you want, then this looks ok. If you think it through and decide some more precise(?) check is needed, then I guess that would be a requested change.

globus_automate_client/flows_client.py
```diff
@@ -460,25 +463,35 @@ def list_flows(
...
-        if filters is not None:
-            params.update(filters)
+
         if orderings:
-            builder = []
-            for field, value in orderings.items():
-                builder.append(f"{field} {value}")
-            params["orderby"] = ",".join(builder)
+            params["orderby"] = ",".join(
+                f"{field} {value}" for field, value in orderings.items()
```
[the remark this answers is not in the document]
kurtmckee: Haha, thanks! :smile:

globus_automate_client/flows_client.py
kurtmckee: In the specific places where I used it, truthiness appears to be correct as the goal is to eliminate empty values like empty strings and empty lists.

It's true that the user could pass in arbitrary types and values in violation of the type annotations, but the original `if x is not None` checks wouldn't have caught or prevented that, either. For example, the user could pass in `slice(2, 7)` and potentially cause a TypeError during JSON encoding.

Please let me know if that's satisfactory reasoning, or if more discussion is required. I might be missing something you're trying to gently lead me to, as well! :grin:

kurtmckee:
@urimandujano, I reworded the commit messages to include references to Shortcut stories.

Please let me know if there's anything else needed here. Thanks!

globus_automate_client/flows_client.py
jimPruyne: yep, that's fine. Just something I always watch out for, and if your thinking is that the various truthiness cases are what we want, I'm good. And, I agree, the `None` check had its own set of possible gotchas.

jimPruyne (approved):

rudyardrichter (approved):

PR #99, opened 2021-11-05
Upgrade to SDK v3

kurtmckee:
This upgrades to SDK v3. The unit tests PR will follow after this is merged, as they depend on this PR.

Features
--------

-   Upgrade to Globus SDK v3.


globus_automate_client/cli/auth.py
```diff
@@ -296,8 +296,6 @@ def refresh_handler(
                     expires_at=token_set.expiration_time,
                     on_refresh=refresh_handler,
                 )
```
urimandujano, line 298: I think we were doing this to force the authorizer to refresh its token if it was expired. It looks like this got renamed and is getting called implicitly in `get_authorization_header`. So it sounds like we don't actually need to do this.

@jimPruyne can you remember a specific reason we were doing this ourselves? If not, I'm happy removing this and trusting the RefreshingAuthorizer to do its thing.

globus_automate_client/queues_client.py
```diff
@@ -130,17 +133,17 @@ def receive_messages(
...
-        return self.get(f"/queues/{queue_id}/messages", params=params)
+        return self.get(f"/queues/{queue_id}/messages", query_params=params)
 
     def delete_messages(
         self, queue_id: str, receipt_handles: List[str]
     ) -> GlobusHTTPResponse:
         self.authorizer = get_authorizer_for_scope(QUEUES_RECEIVE_SCOPE)
         body = {"data": [{"receipt_handle": rh} for rh in receipt_handles]}
-        return self._request(
+        return self.request(
```
urimandujano, line 143: Would it be more consistent to use the client's delete method? Based on how the rest of it looks we could do `self.delete(f"/queues/{queue_id}/messages", data=body)`

pyproject.toml
```diff
@@ -62,6 +62,8 @@ mypy = "^0.910"
 types-PyYAML = "^5.4.6"
 types-requests = "^2.25.6"
 scriv = {extras = ["toml"], version = "^0.12.0"}
+tox = "^3.24.4"
+pre-commit = "^2.15.0"
```
urimandujano, line 66: Should we add these in the follow up PR (I think you mentioned the next one would be adding tests?)

globus_automate_client/action_client.py
```diff
@@ -176,22 +180,20 @@ def log(
...
             params["per_page"] = per_page
-        path = self.qjoin_path(action_id, "log")
-        return self.get(path, params=params)
+        return self.get(f"{quote(action_id)}/log", query_params=params)
 
     @classmethod
     def new_client(
         cls: Type[_ActionClient],
         action_url: str,
         authorizer: AllowedAuthorizersType,
```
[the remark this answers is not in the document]
urimandujano: Sorry GH really messed this comment up. My comment is for the line immediately below where we drop the http_timeout parameter.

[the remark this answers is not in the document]
kurtmckee: ARGH nope, the *http_timeout* parameter disappeared but I didn't dig deep enough. *http_timeout* doesn't exist anymore, but if it's passed in a dictionary named *transport_params* it'll make its way to the right place.

I'll add this back in.

globus_automate_client/client_helpers.py
```diff
@@ -126,5 +123,4 @@ def create_flows_client(
         base_url=base_url,
         authorizer=authorizer,
         authorizer_callback=authorizer_callback,
-        http_timeout=http_timeout,
```
kurtmckee, line 129: This needs to be added back in based on @urimandujano's observation above.

globus_automate_client/flows_client.py
```diff
@@ -1002,7 +1003,6 @@ def new_client(
         authorizer_callback: AuthorizerCallbackType,
         authorizer: AllowedAuthorizersType,
         base_url: Optional[str] = None,
-        http_timeout: int = 10,
```
kurtmckee, line 1005: Add this back in.

globus_automate_client/queues_client.py
kurtmckee: The client's delete method doesn't have a *data* parameter.

pyproject.toml
kurtmckee: It's possible. BUT! I changed these lines in the same commit like a genius and didn't want to rip apart the commit at that level (and regenerate `poetry.lock`) to remove these. As they're dev dependencies that have no config files triggering them yet I would like to leave them in here. Let me know if that's acceptable in this situation.

globus_automate_client/action_client.py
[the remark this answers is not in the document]
kurtmckee: Should be fixed now. Please double-check. =)

urimandujano:
Thanks for the updates Kurt! I'm on board with your comments about the dev-dependencies (I appreciate all the git magic you've already done just to get this PR together!). This looks good to me but I'd like to hold off until @jimPruyne can take a look at my comment above since he implemented that piece and has a better understanding of its behavior.

urimandujano:
@rudyardrichter / @jimPruyne can either of you comment on if we need to or should continue to call `check_expiration_time()` on the `RefreshAuthorizer`? I'm @ Rudy since he works on the globus-sdk and I also see [the timer-cli making the same check](https://github.com/globusonline/timer_cli/blob/4b67efa91fea10e55adb618d96c97b6c3604ec16/timer_cli/auth.py#L226).

kurtmckee:
Resolved merge conflicts. @rudyardrichter / @jimPruyne, when you're able to comment on https://github.com/globus/globus-automate-client/pull/99#issuecomment-970643901 I think that will help drive this PR to closure. Thanks!

globus_automate_client/cli/auth.py
rudyardrichter: This is correct; `.get_authorization_header` should handle this.

rudyardrichter:
I'm taking a look at this now.

globus_automate_client/cli/auth.py
rudyardrichter: In fact, with the 3.0 additions, if you have an authorizer attached to your `BaseClient `implementation I believe none of that is even necessary; `RequestsTransport.request` is handling it all.

globus_automate_client/action_client.py
```diff
@@ -25,8 +26,11 @@ class ActionClient(BaseClient):
         AccessTokenAuthorizer, RefreshTokenAuthorizer, ClientCredentialsAuthorizer
     ]
 
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
+    base_path: str = ""
+    service_name: str = "actions"
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
```
rudyardrichter, line 33: No-op init—could be removed?

rudyardrichter:
This is looking good. I think there's probably room to trim some archaic auth code, but not necessarily in the scope of the PR. The weird `self.authorizer = ...` overwrites in the flows client seem to suggest we need a `FlowClient`, singular, to deal with those cases instead of monkeypatching the `FlowsClient` which should really only be using the `flow_management_authorizer`.

It's a large can of worms, but I think we should next look to stripping out the entire `TokenCache` stuff, because the SDK's `StorageAdapter` should be able to handle that.

globus_automate_client/cli/auth.py
kurtmckee: Thanks for the review, Rudy! To confirm: you're saying that removing this call is okay? Or is there something I need to address here?

globus_automate_client/action_client.py
kurtmckee: Good catch! Will do.

kurtmckee:
The reassignments to `self.authorizer` will disappear in the next PR, which is the final piece: the introduction of unit tests.

globus_automate_client/action_client.py
```diff
@@ -176,15 +180,14 @@ def log(
         # *reverse_order* MUST BE None to prevent reversing the sort order.
         # Any other value, including False, will reverse the sort order.
```
jimPruyne, line 181: Is this comment still true? Was it ever?

globus_automate_client/cli/auth.py
jimPruyne: The original purpose was, I believe, as @urimandujano says. If that is a changed behavior in SDK3, as @rudyardrichter says, I think I agree with you @kurtmckee , but I would also need clarification.

globus_automate_client/flows_client.py
```diff
@@ -491,7 +492,7 @@ def list_flows(
             filters.pop("per_page", None)
             params.update(filters)
 
-        return self.get("/flows", params=params, **kwargs)
+        return self.get("/flows", query_params=params, **kwargs)
```
jimPruyne, line 495: Is `params` -> `query_params` an SDK3 change? I wouldn't have expected that change as it doesn't seem to add much (and maybe changes from what `requests` does which would have been a reasonable thing to follow?)

globus_automate_client/queues_client.py
```diff
@@ -81,19 +84,19 @@ def update_queue(
...
-        body = dict(
-            id=queue_id,
-            label=label,
-            admins=admins,
-            senders=senders,
-            receivers=receivers,
-            delivery_timeout=delivery_timeout,
-            visibility_timeout=visibility_timeout,
-        )
+        body = {
```
jimPruyne, line 87: This is truly the more common method, but does anyone else kinda like the `dict(**kwargs)` syntax?

globus_automate_client/queues_client.py
```diff
@@ -130,17 +133,17 @@ def receive_messages(
...
         if receive_request_attempt_id is not None:
             params["receive_request_attempt_id"] = receive_request_attempt_id
-        return self.get(f"/queues/{queue_id}/messages", params=params)
+        return self.get(f"/queues/{queue_id}/messages", query_params=params)
 
     def delete_messages(
         self, queue_id: str, receipt_handles: List[str]
     ) -> GlobusHTTPResponse:
         self.authorizer = get_authorizer_for_scope(QUEUES_RECEIVE_SCOPE)
         body = {"data": [{"receipt_handle": rh} for rh in receipt_handles]}
```
jimPruyne, line 142: I guess the convention of calling this `data` elsewhere had kind of been put in place...

tests/test_sdk_v3.py
```diff
@@ -0,0 +1,5 @@
+import globus_automate_client
+
+
+def test_import():
+    assert bool(globus_automate_client)
```
jimPruyne, line 5: This is actually kind of a cool test in a sort of reductive way.

jimPruyne (approved):
I don't think I have any particularly critical _new_ comments. As long as all comments that arose before I got here are resolved, I'm good with this.

globus_automate_client/action_client.py
kurtmckee: Yes, it is true. While developing #90 I found that `reverse_order` was extremely sensitive: `False` and `True` both resulted in a reversed sort order, perhaps because the values were interpreted as strings by the server. To work around this, the client was setting a default of `None` so that if the user didn't specify a reverse order, that default `None` would make its way in as the value that's passed here.

In #90 I adjusted the code so that it internally uses booleans for `reverse_order` everywhere and swaps out `False` for `None` at this point in the code. This comment is here to clarify why this value swap is happening.

globus_automate_client/flows_client.py
kurtmckee: This changed throughout the SDK's request methods. See [globus_sdk.client.BaseClient.get()](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/client.py#L120).

tests/test_sdk_v3.py
kurtmckee: Haha, very! It was an early best-effort attempt to establish that, at the very least, it was possible to import the SDK. The unit test PR will flesh this out significantly. =)

globus_automate_client/queues_client.py
kurtmckee: It seems to vary between methods; I saw `req_body`, `data`, `body`, and `payload` in other places but didn't standardize this throughout.

globus_automate_client/queues_client.py
kurtmckee: I prefer dict literal syntax, though like all preferences this is just due to acclimation. :smiley:

globus_automate_client/cli/auth.py
kurtmckee: I may be reading this incorrectly, but it looks like the expiration time is checked here, in the `ensure_valid_token()` method. This is indeed called from within `RenewingAuthorizer.get_authorization_header()`.

https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/authorizers/renewing.py#L138

It's possible to call `ensure_valid_token()` explicitly. Does that need to happen here?

rudyardrichter: Kurt has it. I can link more, specific pieces that tie into this. (The links are to my fork, sorry; just easier to generate from my editor.)

- `BaseClient.request` (used for `BaseClient.get` et al) [uses this intermediate `RequestsTransport.request`](https://github.com/rudyardrichter/globus-sdk-python/blob/a346def64ce6b1490ea38aa62b92f884f01552bd/src/globus_sdk/client.py#L282).
- `RequestsTransport.request` [calls `RequestsTransport._set_authz_header`](https://github.com/rudyardrichter/globus-sdk-python/blob/a346def64ce6b1490ea38aa62b92f884f01552bd/src/globus_sdk/transport/requests.py#L231)
- `RequestsTransport._set_authz_header` [calls `GlobusAuthorizer.get_authorization_header`](https://github.com/rudyardrichter/globus-sdk-python/blob/a346def64ce6b1490ea38aa62b92f884f01552bd/src/globus_sdk/transport/requests.py#L174)
- ...which finally, **if you are using a `RenewingAuthorizer` and not overriding `get_authorization_header`,** makes sure the token works.

PR #101, opened 2021-11-10
Removes duplicate entries from CHANGELOG

urimandujano:
After working through the release for 0.12.3 I noticed that `CHANGELOG.rst` had duplicate entries. This is because I ran `poetry run scriv collect --keep --version TEST` previously and had a modified copy of the changelog locally. This fix removes those entries to try to keep the changelog clean.

@kurtmckee let me know if you think this is a good way to handle this particular fix.


kurtmckee:
Deleting the duplicate changelog text is definitely the right way to handle this. I wouldn't include a changelog fragment for this one. Still approving, but I would recommend removing the changelog fragment for this change.

kurtmckee (approved):
I recommend removing the changelog fragment but I approve this in general.

urimandujano:
Ok I went with removing the fragment since it did feel like something that didn't need announcing. Thanks for your input

PR #102, opened 2021-11-12
Document the HTTP action provider

kurtmckee:
This branch adds documentation for the upcoming HTTP action provider, including:

* The URL and scope
* Complete documentation of all input document keys and values
* Examples of usage
* Documentation of the errors that may be returned by the HTTP action provider, including suggestions for how to resolve the errors.

[sc-11704]

### DIFF
--- a/docs/source/globus_action_providers.rst
+++ b/docs/source/globus_action_providers.rst
@@ -617,15 +617,15 @@ For example, the URL may require an ID to be embedded in the path:
 
     https://domain.example/dataset/37/status/42/info
 
-When this is required you can specify substitution patterns in the ``url``
+When this is required you can specify substitution identifiers in the ``url``
 and dynamically provide the correct values in the input document.
 
-Substitution patterns in the URL are wrapped in curly braces like "``{id}``".
-You can specify multiple substitution patterns, and patterns can be repeated.
-Each pattern MUST be referenced as a key in the ``url-substitutions`` object.
+Substitution identifiers in the URL are wrapped in curly braces like "``{id}``".
+You can specify multiple substitution identifiers, and identifiers can be repeated.
+Each identifier MUST be referenced as a key in the ``url-substitutions`` object.
 
-If patterns are specified in the URL but are not provided in the ``url-substitutions`` object an error will be returned.
-Similarly, if pattern values are provided in ``url-substitutions`` but are not required then this will also be treated as an error.
+If identifiers are specified in the URL but are not provided in the ``url-substitutions`` object an error will be returned.
+Similarly, if identifier values are provided in ``url-substitutions`` but are not required then this will also be treated as an error.
 
 The input document below will result in the same URL as the example above,
 but the dataset ID and the status ID can now be dynamically specified during Flow execution.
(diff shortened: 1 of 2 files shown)


docs/source/globus_action_providers.rst
```diff
@@ -408,3 +408,371 @@ should be used in the Datacite test service or the production service.
...
+*   Content submission
+*   URL customization
+
+..  note::
+
+    The JSON schema that authoritatively describes the accepted keys and values can be retrieved using this command:
+
+    ..  code-block:: text
+
+        globus-automate action introspect --action-url https://actions.globus.org/http
```
urimandujano, line 448: Since this isn't specific to this AP I think it makes sense to avoid duplication in favor of what's at the top of the file. It could be that the text at the top of the file could use reformatting/updating to make it more clear that it applies to every listed AP.

docs/source/globus_action_providers.rst
```diff
@@ -408,3 +408,371 @@ should be used in the Datacite test service or the production service.
...
+
+Errors: JSON schema validation
+..............................
+
+Input documents will be validated against the HTTP action provider JSON schema.
+You can get an authoritative copy of the JSON schema by running this command:
+
+..  code-block:: text
+
+    globus-automate action introspect --action-url https://actions.globus.org/http
```
urimandujano, line 677: Same idea here as above. Although this one is tricky since its dealing specifically with debugging errors with JSON validation. It may be valuable to add this into its own section since this is useful information that applies to all APs and their behavior will more or less be the same.

urimandujano (approved):
These docs read really well and set the goal for where we want all our AP documentation to get to. I'd personally use this as a template to fill in the details on the other APs. My comments are opinions on reducing duplicated information but if you feel strongly about keeping as is I'm ok with that.

docs/source/globus_action_providers.rst
kurtmckee: 100% agree. Should that be in the scope of this work? If not I'll create a story.

docs/source/globus_action_providers.rst
kurtmckee: Agree here, too. As above, if this needs to be handled in this PR I'll make the changes but this may be useful in the context of a broader doc update.

docs/source/globus_action_providers.rst
urimandujano: If you have any immediate thoughts on how to make the text on introspection [at the top of the doc here](https://github.com/globus/globus-automate-client/blob/master/docs/source/globus_action_providers.rst) more visible then I'd say we can make the changes here on this PR. If not then let's cut the duplicated instructions here and keep an eye out for a potential doc update in the future. From my experience doc work can be a deep rabbit hole and I don't think a broad update is necessary right now.

kurtmckee: Oh! I just realized that you're pointing out that the instructions are already documented at the top of the doc. :disappointed_relieved: Okay, I'll get this updated.

docs/source/globus_action_providers.rst
kurtmckee: Done.

docs/source/globus_action_providers.rst
kurtmckee: Done. Thanks for pointing this out!

kurtmckee:
Note that the changes in b6defdd depend on https://github.com/globusonline/globus-automate/pull/218.

urimandujano (commented):
Thanks Kurt! Thanks for mentioning that this relies on #218 -- let's hold off on merging this to master until we cut that PR in a release, otherwise we'll be out of sync with what is actually deployed.

kurtmckee:
Updated to remove references to the user-configurable *timeout* key (which is no longer supported) and to document the new HTTP response size and total timeout errors.

kurtmckee:
The HTTP action provider is currently for internal use. Once this restriction is lifted this can be reviewed and merged.

PR #103, opened 2021-11-16
Allows empty strings and lists in update_flow

urimandujano:
This PR adds logic to the `FlowsClient`'s `update_flow` method to differentiate between updates that set fields to empty objects and updates that are not setting fields at all. Fields now default to `None` and if they are set to non-`None` values they are carried over into the API request. What this allows the SDK to do is to empty out role-based access control (RBAC) lists and set text fields to empty strings (note that the API technically doesn't disallow empty strings).


kurtmckee (approved):
Looks good to me! I'll make sure that this ripples into the SDK v3 and unit test branches when this is merged.

jimPruyne (approved):
truthiness strikes again

PR #104, opened 2021-11-19
Add testing and linting

kurtmckee:
This is the final PR of the original SDK v3 branch. It adds unit testing, linting, and pre-commit hooks. The tests are not yet comprehensive but increase code and branch coverage to nearly 100% for the files they test.

@rudyardrichter I rebased many of the original incremental commits into much broader batches (like grouping most of the file-by-file tests together). Please let me know if this would benefit from further rebasing.

# Changelog

Bugfixes
--------

-   Fix a bug that could allow the Flows authorizer to be lost if an exception was raised.
    (Authorizer swaps are now handled using a context manager.)

Other
-----

-   Add code linting, documentation build testing, and a bunch of unit tests.
-   Add GitHub Actions to run on push and pull requests.
-   Add a pre-commit configuration file to increase overall code quality.

-----

# NOTE

After this change is merged, all developers will need to run the following commands to update their environment:

```
poetry install
pre-commit install
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,9 +18,3 @@ repos:
   hooks:
     - id: isort
       name: isort (python)
-    - id: isort
-      name: isort (cython)
-      types: [cython]
-    - id: isort
-      name: isort (pyi)
-      types: [pyi]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,26 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v3.2.0
+  hooks:
+  - id: trailing-whitespace
+  - id: end-of-file-fixer
+  - id: check-yaml
+  - id: check-added-large-files
+- repo: https://github.com/psf/black
+  rev: 21.9b0
+  hooks:
+  - id: black
+    language_version: python3
+- repo: https://github.com/pycqa/isort
+  rev: 5.8.0
+  hooks:
+    - id: isort
+      name: isort (python)
+    - id: isort
+      name: isort (cython)
+      types: [cython]
+    - id: isort
+      name: isort (pyi)
+      types: [pyi]
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -84,8 +84,9 @@ class ActionClient(BaseClient):
         :param force_path: A URL to use for running this action, ignoring any
             previous configuration
         :param label: Set a label for the Action that is run.
-        :param run_monitors: May be used as an alias for ``monitor_by``
-        :param run_managers: May be used as an alias for ``manage_by``
+
+        ``run_monitors`` may be used as an alias for ``monitor_by``,
+        and ``run_managers`` may be used as an alias for ``manage_by``.
         """
         if request_id is None:
             request_id = str(uuid.uuid4())
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -176,7 +176,7 @@ class ActionClient(BaseClient):
 
         # *reverse_order* MUST BE None to prevent reversing the sort order.
         # Any other value, including False, will reverse the sort order.
-        params: Dict[str, Union[int, str]] = {
+        params: Dict[str, Union[int, str, bool, None]] = {
             "reverse_order": True if reverse_order else None,
             "limit": limit,
         }
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -29,6 +29,10 @@ class ActionClient(BaseClient):
     base_path: str = ""
     service_name: str = "actions"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._action_scope: Optional[str] = None
+
     @property
     def action_scope(self) -> str:
         """
@@ -39,7 +43,7 @@ class ActionClient(BaseClient):
         have to have been provided on initialization to the ``ActionClient``.
         Otherwise, this call will fail.
         """
-        if not hasattr(self, "_action_scope"):
+        if self._action_scope is None:
             resp = self.introspect()
             if resp.data is None:
                 self._action_scope = ""
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -2,13 +2,8 @@ import uuid
 from typing import Any, Dict, Iterable, Mapping, Optional, Type, TypeVar, Union
 from urllib.parse import quote
 
-from globus_sdk import (
-    AccessTokenAuthorizer,
-    BaseClient,
-    ClientCredentialsAuthorizer,
-    GlobusHTTPResponse,
-    RefreshTokenAuthorizer,
-)
+from globus_sdk import BaseClient, GlobusHTTPResponse
+from globus_sdk.authorizers import GlobusAuthorizer
 
 from .helpers import merge_lists
 
@@ -16,16 +11,6 @@ _ActionClient = TypeVar("_ActionClient", bound="ActionClient")
 
 
 class ActionClient(BaseClient):
-    allowed_authorizer_types = (
-        AccessTokenAuthorizer,
-        RefreshTokenAuthorizer,
-        ClientCredentialsAuthorizer,
-    )
-
-    AllowedAuthorizersType = Union[
-        AccessTokenAuthorizer, RefreshTokenAuthorizer, ClientCredentialsAuthorizer
-    ]
-
     base_path: str = ""
     service_name: str = "actions"
 
@@ -195,7 +180,7 @@ class ActionClient(BaseClient):
     def new_client(
         cls: Type[_ActionClient],
         action_url: str,
-        authorizer: AllowedAuthorizersType,
+        authorizer: Optional[GlobusAuthorizer],
         http_timeout: int = 10,
     ) -> _ActionClient:
         """
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -51,7 +51,7 @@ class ActionClient(BaseClient):
                 self._action_scope = resp.data.get("globus_auth_scope", "")
         return self._action_scope
 
-    def introspect(self, **kwargs) -> GlobusHTTPResponse:
+    def introspect(self, **_) -> GlobusHTTPResponse:
         """
         Introspect the details of an Action Provider to discover information
         such as its expected ``action_scope``, its ``input_schema``, and who to
@@ -211,9 +211,12 @@ class ActionClient(BaseClient):
             the Action Provider to be made.
 
         **Examples**
-            >>> authorizer = ...
-            >>> action_url = "https://actions.globus.org/hello_world"
-            >>> ac = ActionClient.new_client(action_url, authorizer)
+
+        ..  code-block:: pycon
+
+            >>> auth = ...
+            >>> url = "https://actions.globus.org/hello_world"
+            >>> ac = ActionClient.new_client(url, auth)
             >>> print(ac.run({"echo_string": "Hello from SDK"}))
         """
         return cls(
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -4,11 +4,11 @@ from urllib.parse import quote
 
 from globus_sdk import (
     AccessTokenAuthorizer,
+    BaseClient,
     ClientCredentialsAuthorizer,
     GlobusHTTPResponse,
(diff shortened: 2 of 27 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  rev: 5.8.0
+  hooks:
+    - id: isort
+      name: isort (python)
+    - id: isort
+      name: isort (cython)
+      types: [cython]
+    - id: isort
+      name: isort (pyi)
+      types: [pyi]
```
rudyardrichter, line 26: What's the purpose for these other isort types?

globus_automate_client/cli/auth.py
```diff
@@ -68,12 +72,12 @@ def __init__(self, token_store: Union[pathlib.Path, str]):
         self.modified = False
 
     @property
-    def tokens_for_environment(self) -> Dict[str, TokenSet]:
+    def tokens_for_environment(self):
```
rudyardrichter, line 75: Why is the annotation gone?

globus_automate_client/cli/auth.py
```diff
@@ -138,7 +142,7 @@ def load_tokens(self):
             )
 
     @staticmethod
-    def _make_jsonable(tokens: TokensInTokenCache) -> Dict[str, Any]:
+    def _make_jsonable(tokens) -> Dict[str, Any]:
```
rudyardrichter, line 145: Same here: checking why annotation removed

globus_automate_client/cli/constants.py
```diff
@@ -24,10 +23,18 @@ class FlowRoleDeprecated(str, enum.Enum):
     administered_by = "administered_by"
 
 
-# Adapted from https://stackoverflow.com/questions/33679930/how-to-extend-python-enum
-FlowRoleAllNames = enum.Enum(
-    "FlowRoleAllNames", [(i.name, i.value) for i in chain(FlowRole, FlowRoleDeprecated)]
-)
+class FlowRoleAllNames(str, enum.Enum):
```
rudyardrichter, line 26: I wonder if this could be left as the original version, and we use this:
```python
from typing import TypeVar, Union

T = TypeVar("T", bound=Union[FlowRole, FlowRoleDeprecated, FlowRoleAllNames])
```
That might work? Is mypy appeased?

globus_automate_client/flows_client.py
```diff
@@ -285,31 +304,32 @@ def deploy_flow(
...
-        )
-        temp_body["flow_starters"] = merge_lists(
-            flow_starters, kwargs, "runnable_by", "starters"
-        )
-        temp_body["flow_administrators"] = merge_lists(
-            flow_administrators, kwargs, "administered_by", "administrators"
-        )
-        temp_body["subscription_id"] = subscription_id
-        # Remove None / empty list items from the temp_body
+        temp_body: Dict[str, Any] = {
```
rudyardrichter, line 315: Wow, this is massively more readable to me. Thanks for all this miscellaneous cleanup!

tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+import responses
+
+
+@pytest.fixture
+def mocked_responses():
```
rudyardrichter, line 6: Maybe this could be set to `autouse=True`? I don't know offhand whether that disrupts some specifictests, but I imagine we probably just want this on all the time.

tests/test_init.py
```diff
@@ -0,0 +1,18 @@
+import inspect
+
+import globus_automate_client
+
+
+def test_all_exports():
```
rudyardrichter, line 6: This test seems like a tautology? In which cases could something appear in `__all__` yet not appear under the import or vice versa?

rudyardrichter:
As a PR-meta-comment, similar to on other PRs in this recent effort, I'd vote in favor of some (fairly aggressive) commit squashing.

.pre-commit-config.yaml
kurtmckee: Those were copy-pasted from isort's documentation. I don't think they contribute anything, and they can be removed if needed.

globus_automate_client/cli/auth.py
kurtmckee: Both of these lines were removed in 2f7bf211d6320fa05f3b64b1d06e2a01e40aacd4 because mypy _super_ dislikes these lines. No matter how specific I got, no matter how I changed it, mypy traced problems deeper into the code and to other usages.

Removing the lines resolved the problem in an incorrect way.

globus_automate_client/cli/auth.py
kurtmckee: See answer above. You caught both lines from 2f7bf211d6320fa05f3b64b1d06e2a01e40aacd4 that are, to sum it up, incorrect fixes.

globus_automate_client/cli/constants.py
kurtmckee: It can't be left as-is -- mypy's complaint is not the types but rather that dynamic enums fundamentally break the concept of having enums.

globus_automate_client/flows_client.py
kurtmckee: My pleasure!

tests/conftest.py
kurtmckee: You're right, this should probably be on all the time to prevent unexpected I/O. I'll try that out and push an update or report back.

tests/test_init.py
kurtmckee: `__all__` isn't checked by Python unless you explicitly use `from blah import *`. Therefore this test ensures that future development both _imports_ what's required by `__all__` and _exports_ what was almost certainly imported for the purpose. For example:


```python
from submodule1 import ExistingThing
from submodule2 import KurtImportedMe

__all__ = (
    "ExistingThing",
    "KurtImportedMeButCannotCopyAndPasteCorrectly",
)
```

You can import that file using `import blah` without any issue, but you'll get an unexpected `AttributeError` if you use `from blah import *`.

globus_automate_client/cli/auth.py
kurtmckee: I'm definitely open to revisiting this, but it will take some time to unravel the threads that mypy tugged at when I first tried fixing this...

tests/conftest.py
kurtmckee: Done. Thanks for this suggestion!

.pre-commit-config.yaml
kurtmckee: These are removed now. Thanks for catching this!

tests/test_init.py
kurtmckee: I'm going to mark this as resolved but please reopen as needed!

kurtmckee:
@rudyardrichter would it be acceptable to do a squash-merge at the end? I can also rebase this with a heavier hand.

globus_automate_client/cli/auth.py
kurtmckee: I tried to fix this again just now but mypy really doesn't like these annotations. I'm open to suggestions how to proceed!

globus_automate_client/action_client.py
```diff
@@ -84,8 +88,9 @@ def run(
         :param force_path: A URL to use for running this action, ignoring any
             previous configuration
         :param label: Set a label for the Action that is run.
-        :param run_monitors: May be used as an alias for ``monitor_by``
-        :param run_managers: May be used as an alias for ``manage_by``
+
+        ``run_monitors`` may be used as an alias for ``monitor_by``,
+        and ``run_managers`` may be used as an alias for ``manage_by``.
```
urimandujano, line 93: These methods are actually documented on our client doc site and the `:param:` format for the docstring is how Sphinx knows to parse these and render them on readthedocs.

Let's keep the pre-existing to make sure our docs stay in sync with the code.

globus_automate_client/action_client.py
```diff
@@ -206,9 +211,12 @@ def new_client(
             the Action Provider to be made.
 
         **Examples**
-            >>> authorizer = ...
-            >>> action_url = "https://actions.globus.org/hello_world"
-            >>> ac = ActionClient.new_client(action_url, authorizer)
+
+        ..  code-block:: pycon
```
urimandujano, line 215: Is this how you get python syntax highlighting? This is awesome!

globus_automate_client/flows_client.py
```diff
@@ -227,9 +227,28 @@ def __init__(
     ) -> None:
         super().__init__(**kwargs)
         self.client_id = client_id
-        self.flow_management_authorizer: AllowedAuthorizersType = self.authorizer
         self.get_authorizer_callback = get_authorizer_callback
 
+    @contextlib.contextmanager
+    def use_temporary_authorizer(self, authorizer):
```
urimandujano, line 233: Great idea!

globus_automate_client/flows_client.py
```diff
@@ -792,10 +815,9 @@ def enumerate_runs(
             filters.pop("per_page", None)
             params.update(filters)
 
-        self.authorizer = self._get_authorizer_for_flow("", RUN_STATUS_SCOPE, kwargs)
-        response = self.get("/runs", query_params=params, **kwargs)
-        self.authorizer = self.flow_management_authorizer
-        return response
+        authorizer = self._get_authorizer_for_flow("", RUN_STATUS_SCOPE, kwargs)
+        with self.use_temporary_authorizer(authorizer):
+            return self.get("/runs", query_params=params, **kwargs)
```
urimandujano, line 820: 👌🏼

tests/test_helpers.py
```diff
@@ -0,0 +1,59 @@
...
+        # Multiple values
+        (
+            ([1, 2], [2, 3], {"k1": [4], "k2": [5], "k3": [6]}, "k2", "k3", "bogus"),
+            {1, 2, 3, 5, 6},
+            "all values must be found",
+        ),
+    ),
+)
+def test_merge_lists(args, expected, message):
+    """Validate globus_automate_client.helpers.merge_lists()."""
```
urimandujano, line 31: This was what I was missing as I looked at `sc-11808`.  Did you happen to run into any cases that were problematic? I still haven't had a chance to investigate/patch that bug but if that function made sense to you I would appreciate an extra set of eyes on that.

urimandujano (changes requested):
I'm a little sad that we had to remove the annotations on our token handling code to get mypy passing but that reinforces our decision to make moving onto the globus-sdk v3's bulit in storage solution. We'll probably get more accurate type coverage from making the switch.

globus_automate_client/action_client.py
kurtmckee: I understand that the `:param:` format is used by Sphinx and autodoc. I deliberately changed these because neither are actual parameters anymore and appeared to be deprecated in favor of `monitor_by` and `manage_by`. I wanted to reduce their visibility so new code is less likely to depend on these fallback aliases.

I'll move it back.

kurtmckee: Done.

tests/test_helpers.py
kurtmckee: Sure thing! I'll dig into that. I'll mark this as resolved and look into `sc-11808`.

globus_automate_client/cli/auth.py
rudyardrichter: I see quite a lot of mypy errors either way—I'm not sure if maybe I don't have this set up right? As is maybe we can leave the annotation, and leave resolving the errors to a future effort on typing.

kurtmckee: @rudyardrichter I'm now able to reproduce the mypy errors. This may point to some issues between SDK v3 and the client unfortunately, because some of the errors are "such-and-such attribute doesn't exist on that class". I'll keep working on this.

globus_automate_client/action_client.py
urimandujano: Oh I see. The deprecated behavior is actually reverse -- the new terminology is `run_monitors` and `run_managers` and we're trying to migrate folks off of `monitor_by` and `manage_by`. Sorry for causing some flipflop here.

urimandujano (approved):

jimPruyne (approved):
Looking good to me. A couple bummers where annotations had to get dropped, but I really like having the GH Action for testing and the pre-commit!

Definitely a huge step in the right direction.

globus_automate_client/cli/auth.py
kurtmckee: I think I've resolved all of these issues now. Unit tests are now passing.

globus_automate_client/cli/constants.py
kurtmckee: I'm marking this as resolved. Please re-open if this needs to be addressed differently.

kurtmckee:
The latest push fixes a merge conflict, updates the tests for `merge_keywords()`, and fixes a bug in same.

rudyardrichter:
LGTM!

PR #105, opened 2021-11-23
Allow more value types when merging keyword parameter aliases

kurtmckee:
This change simplifies the code in the `merge_aliases()` function and allows it to merge additional value types, like strings and tuples of strings.

# Changelog

Bugfixes
--------

-   Support strings (and tuples/sets containing strings) as argument values
    when running, deploying, or updating an action or a flow and specifying
    a keyword argument alias like ``visible_to`` or ``runnable_by``.

[sc-11808]

### DIFF
--- a/changelog.d/20211123_092348_kurtmckee_merge_parameter_aliases_better.rst
+++ b/changelog.d/20211123_092348_kurtmckee_merge_parameter_aliases_better.rst
@@ -1,0 +1,6 @@
+Bugfixes
+--------
+
+-   Support strings (and tuples/sets containing strings) as argument values
+    when running, deploying, or updating an action or a flow and specifying
+    a keyword argument alias like ``visible_to`` or ``runnable_by``.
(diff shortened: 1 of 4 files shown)


kurtmckee:
Discovered a typo in a comment and fixed/force-pushed it.

```diff
- # .pop() is destructive operation
+ # .pop() is a destructive operation
```

urimandujano (approved):
Thank you! This looks much more digestible to me so I'm happy with merging this. If there are more issues with functions that use `merge_keywords` we can make sure to track what the inputs values were exactly that caused 11808.

jimPruyne (approved):
Don't think I was requested as reviewer on this, but it also gets the :+1: from me.

kurtmckee:
Thanks @jimPruyne! I was trying to restrict PR's to only two reviewers requested so that I'm not tagging everybody on every commit. I'll ask offline how best to handle that so I'm not inappropriately leaving you (or anyone else) off as reviewers.

kurtmckee:
I discovered a typo in the docstring. It's fixed now to reflect the current function name.

PR #108, opened 2021-12-02
Fix a KeyError crash that occurs when `-v` is passed

kurtmckee:
[sc-12117]

### DIFF
--- a/globus_automate_client/cli/helpers.py
+++ b/globus_automate_client/cli/helpers.py
@@ -1,6 +1,7 @@
 import json
 from typing import Any, Callable, Dict, List, Mapping, Optional, Union
 
+import requests
 import typer
 import yaml
 from globus_sdk import GlobusAPIError, GlobusHTTPResponse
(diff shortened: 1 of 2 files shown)


urimandujano (commented):
Tried this out locally and this does patch that bug up. It's too bad that the `GlobusHTTPResponse` object doesn't expose the underlying request data anymore -- I can't remember if that was a private interface anymore. 

I added @rudyardrichter as a reviwer to get his perspective on the SDK usage just in case there's another way to accomplish this.

globus_automate_client/cli/helpers.py
```diff
@@ -37,8 +38,15 @@
...
-        response_status_code = result.data["status_code"]
+        if isinstance(result._response, requests.Response):
+            base_request = result._response.request
+        elif (
+            result._wrapped and isinstance(result._wrapped._response, requests.Response)
+        ):
+            base_request = result._wrapped._response.request
+        else:
+            return "HTTP details are unavailable"
+        response_status_code = result.http_status
```
rudyardrichter, line 49: Nice, I was thinking I might show up here and be able to suggest swapping to `http_status` or something else to that effect. I think this change looks good. Would it be helpful to think about a "debug output" mode on `GlobusHTTPResponse` for cases like this? I feel like `str(response)` being just the JSON could maybe be more helpful, and do something more like this function.

rudyardrichter:
@urimandujano I don't think this is missing out on any SDK tricks or anything, but we could totally add something to this effect on `GlobusHTTPResponse`.

globus_automate_client/cli/helpers.py
kurtmckee: Thanks! Is the "debug output" something to change in this repo or in the SDK? If it's in the SDK I'd like to pass that feedback to the SDK team.

rudyardrichter: Yeah, that's what I was getting at. I'll float that in our next SDK catch-up.

rudyardrichter (approved):

urimandujano:
Thanks for the input here @rudyardrichter. I know accessing "private" attributes is a quick way to break things but it's super convenient for consistently parsing API responses at the CLI level. Adding a standard interface into the error/response details would help streamline some of this code. Let us know how that conversation goes!

urimandujano (approved):

globus_automate_client/cli/helpers.py
kurtmckee: Okay, great. I'm marking this as 'resolved' but please re-open as needed.

PR #109, opened 2021-12-07
Allows workflow to be reused

urimandujano:
This PR adds a the `workflow` call event as a trigger which is necessary to allow other Github Action workflows to use the workflow. This is a stop gap solution to share the enforce changelog fragment behavior across our repositories until something more permanent gets set up.

### DIFF
--- a/.github/workflows/has_changelog.yaml
+++ b/.github/workflows/has_changelog.yaml
@@ -1,6 +1,7 @@
 name: has_changelog
 on:
   - pull_request
+  - workflow_call
 
 jobs:
   check_has_news_in_changelog_dir:


rudyardrichter (approved):

rudyardrichter (changes requested):
The first change looks good, but I think maybe this change will also have to modify the `if: ...` line for the case where this isn't a `pull_request` event

kurtmckee (commented):
I think @rudyardrichter is correct re: modifying the `if:` condition (great catch, Rudy!).

urimandujano:
I thought that was an interesting point since this workflow adds an event (`workflow`) but the parent workflow (like in `action-provider-tools`) will still only trigger on pull_request events. So what happens will depend on how Github exposes the parent workflow's metadata to shared workflows.

I tested this out -- I temporarily updated https://github.com/globus/action-provider-tools/pull/44 to reference the workflow in this PR without changes to the `if...` condition. [Based on the action that ran it looks like the if condition](https://github.com/globus/action-provider-tools/actions/runs/1554843813) still passes because the parent workflow is triggered from a `pull_request`.

So it seems like we can ship this without needing to change anything!

kurtmckee (approved):

PR #117, opened 2021-12-10
Expand testing

kurtmckee:
This adds black testing to the tox test suite -- including for the Jupyter notebooks -- and also adds testing for Python 3.10 (if available on the system). It also switches to use black with Jupyter notebook support in the pre-commit hook.

No additional effort is required to benefit from this, though I updated the dev dependencies so the Jupyter notebook support is available and it will be possible. tox will manage the testing environment and install dependencies for its own testing. Run:

```
tox           # Run the entire test suite start-to-finish
tox -e py310  # Run Python 3.10 only (if available)
tox -e black  # Run black only
```

There isn't a changelog fragment because this is a change that should only affect development, not the product itself.

# Changelog

* Update black pre-commit hook version and add Jupyter notebook support (no developer work required to benefit from this)
* Add `black[jupyter]` to the tox test suite (no developer work required to benefit from this)
* Add `black[jupyter]` as a dev dependency (developers must run `poetry install` to benefit from this)


.pre-commit-config.yaml
```diff
@@ -8,11 +8,13 @@ repos:
   - id: end-of-file-fixer
   - id: check-yaml
   - id: check-added-large-files
+
 - repo: https://github.com/psf/black
-  rev: 21.9b0
+  rev: 21.12b0
   hooks:
-  - id: black
+  - id: black-jupyter
```
urimandujano, line 15: If we switch to this hook, will this still run black as it normally would at commit time?

tox.ini
```diff
@@ -41,6 +42,14 @@ deps =
 commands =
     {envpython} -m sphinx -aWEqb html docs/source docs/build
 
+[testenv:black]
```
urimandujano, line 45: Is this setup purely for test running/setup in a CI environment? It seems like there might be some conflict between what we have in `pyproject.toml` and what gets installed here.

.pre-commit-config.yaml
kurtmckee: Yep! But with the added bonus that it can check Jupyter notebooks as well.

tox.ini
kurtmckee: I'll have to double-check. tox -- like pre-commit -- sets up isolated virtual environments with the listed dependencies. However, while I explicitly listed the dependence on `black` I also said to use `black` from within the existing environment so I need to double-check the outcome here, and I enabled the `skip_install` option so I need to double-check the outcome.

I'm pretty sure it's not setting up a virtual environment for these yet but I'll double-check and get back to you tomorrow morning. Thanks for noticing this and bringing it up!

kurtmckee: It's very definitely running from within a tox virtual environment but I've made that more explicit by using syntax like this:

```
{envpython} -m black ...
{envpython} -m flake8 ...
{envpython} -m isort ...
{envpython} -m mypy ...
```

I also manually uninstalled black/flake8/isort/mypy and confirmed that tox was still able to successfully run the lint tests.

The end result is that the tests are independent of the developers' current paths, environments, and setup.

urimandujano (approved):

kurtmckee:
@rudyardrichter and @jimPruyne I'm interested in getting your feedback on this PR. Thanks for looking through it when you have an opportunity!

PR #121, opened 2022-01-07
Adds an input schema sample Flow definition

urimandujano:
The Single Transfer Flow definition was missing an input schema. This PR adds its schema.

### DIFF
--- a/examples/flows/single-transfer/input-schema.json
+++ b/examples/flows/single-transfer/input-schema.json
@@ -1,0 +1,31 @@
+{
+    "type": "object",
+    "additionalProperties": false,
+    "required": [
+        "source_endpoint_id",
+        "destination_endpoint_id",
+        "transfer_items"
+    ],
+    "properties": {
+        "source_endpoint_id": {
+            "type": "string"
+        },
+        "destination_endpoint_id": {
+            "type": "string"
+        },
+        "transfer_items": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "source_path": {
+                        "type": "string"
+                    },
+                    "destination_path": {
+                        "type": "string"
+                    }
+                }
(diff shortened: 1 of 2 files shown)


examples/flows/single-transfer/input-schema.json
```diff
@@ -0,0 +1,31 @@
...
+            "items": {
+                "type": "object",
+                "properties": {
+                    "source_path": {
+                        "type": "string"
+                    },
+                    "destination_path": {
+                        "type": "string"
+                    }
+                }
```
kurtmckee, line 27: Is it possible to specify that these properties are required? Are other properties allowed?

urimandujano: Yes and yes! I updated the schema to match the transfer_items schema definition on the transfer/transfer Action Provider. It's now exactly what's defined over there.

examples/flows/single-transfer/input-schema.json
```diff
@@ -0,0 +1,39 @@
...
+                "properties": {
+                    "source_path": {
+                        "type": "string"
+                    },
+                    "destination_path": {
+                        "type": "string"
+                    },
+                    "recursive": {
+                        "type": "boolean"
+                    }
```
kurtmckee, line 34: If this has a default value, would it be okay to add that as:

```js
{
  "type": "boolean",
  "default": true  // or false
}
```

urimandujano: I'd prefer to not add a default value since the Flow itself doesn't define those defaults. It just passes the payload onto the Transfer AP and that's what sets default values.

kurtmckee: Okay great. Thanks for confirming!

kurtmckee (approved):

PR #122, opened 2022-01-11
Flow definition validation via Pydantic

urimandujano:
- Introduces models for defining the relationships and constraints
  for the different state types in Step Functions
- Adds tests validating the models for a given state type
- Re-purposes FlowValidationError to reflect pydantic's
  ValidationError
- Creates a separate FlowSchemaValidationError which is raised when
  validating input schemas.
  
The tests I added also run on the sample Flow definitions in the examples directory. I used those to make sure the validation was working correctly and then the unit tests for each individual Flow state type make sure that the expected failures are triggered.

This is a list of the validation that happens:

### General
- enforces that Parameters with keys ending in JSONPath syntax have values
  beginning with JSONPath syntax
- enforces that Parameters with key values starting in JSONPath syntax have keys
  ending in JSONPath syntax
- enforces boolean fields are either True or False with no boolean
  interpretation
- enforces data types (booleans are booleans, objects are dicts, strings are
  strings, ints are ints)
- prevents bogus fields from getting set on states
- enforces fields that are JSONPaths (such as ResultPath) begin with “$.”

### Flow definition validation
- enforces that startAt is a defined state
- enforces that all “next” states defined in states exist
- enforces that an Action state’s “Catch” conditions refer to an existing state
- enforces that a Choice state’s default condition refers to an existing state
- enforces that a Choice state’s rules contain references to existing states
- enforces that transition states only define a Next or End.
- enforces that all defined states can be reached
- enforces that all state names are between 1 and 128 characters inclusive

### Pass State validation
- enforces that InputPath Result and ResultPath are JSONPaths
- enforces that JSONPath keys in parameters are matched with JSONPath values
- prevents additional properties from getting set

### Wait State validation
- enforces that WaitState’s InputPath, OutputPath, SecondsPath, and
  TimestampPath are JSONPaths
- enforces that exactly one of WaitState’s time fields are set
- prevents additional properties from getting set

### Fail State validation
- prevents additional properties from getting set

### ExpressionEval validation
- enforces that ResultPath is a JSONPath
- enforces that JSONPath keys in parameters are matched with JSONPath values
- prevents additional properties from getting set

### ActionState validation
- enforces that InputPath and ResultPath are JSONPaths
- enforces that JSONPath keys in parameters are matched with JSONPath values
- enforces that exactly one of InputPath or Parameters is set
- enforces that the ActionURL is an HttpUrl
- enforces that a Catcher’s ResultPath is a JSONPath
- enforces that a Catcher’s ErrorEquals list is non-empty
- prevents additional properties from getting set

### ChoiceState validation
- enforces that the Choices list is non-empty
- enforces that an And or Or boolean expression contain at least one rule
- enforces that ChoiceRule path variables are JSONPaths
- enforces that only valid ComparisonOperators are used within ChoiceRules
- enforces that only one ComparisonOperator is used within a ChoiceRule
- enforces that BooleanExpressions contain at least one ChoiceRule
- prevents additional properties from getting set

### DIFF
--- a/globus_automate_client/models.py
+++ b/globus_automate_client/models.py
@@ -32,10 +32,6 @@ def check_jsonpath_syntax_in_dict(item: t.Optional[dict]):
             raise ValueError(
                 f'Key "{k}" indicates its value will be a JSONPath, however its value "{v}" is not a JSONPath'
             )
-        elif not k.endswith(".$") and (isinstance(v, str) and v.startswith("$.")):
-            raise ValueError(
-                f'Value "{v}" is a JSONPath, however its key "{k}" does not indicate a JSONPath'
-            )
         elif isinstance(v, dict):
             check_jsonpath_syntax_in_dict(v)
     return item
@@ -58,7 +54,7 @@ class Catchers(BaseModel):
     _enforce_jsonpath = validator("ResultPath", allow_reuse=True)(enforce_jsonpath)
 
 
-class ChoiceRule(BaseModel):
+class DataTestExpression(BaseModel):
     Variable: str
     StringEquals: t.Optional[str]
     StringEqualsPath: t.Optional[str]
@@ -81,7 +77,7 @@ class ChoiceRule(BaseModel):
     NumericLessThanEqualsPath: t.Optional[str]
     NumericGreaterThanEquals: t.Optional[str]
     NumericGreaterThanEqualsPath: t.Optional[str]
-    BooleanEquals: t.Optional[str]
+    BooleanEquals: t.Optional[StrictBool]
     BooleanEqualsPath: t.Optional[str]
     TimestampEquals: t.Optional[str]
     TimestampEqualsPath: t.Optional[str]
@@ -138,11 +134,32 @@ class ChoiceRule(BaseModel):
         return values
 
 
-class DataTestExpressionChoiceRule(ChoiceRule):
+class DataTestExpressionWithTransitions(DataTestExpression):
     Next: str
 
 
-class BooleanExpressionChoiceRule(BaseModel):
+class BooleanExpression(BaseModel):
+    And: t.Optional[t.List[t.Union[DataTestExpression, "BooleanExpression"]]] = Field(
+        None, min_items=1
+    )
+    Or: t.Optional[t.List[t.Union[DataTestExpression, "BooleanExpression"]]] = Field(
+        None, min_items=1
+    )
+    Not: t.Optional[t.Union[DataTestExpression, "BooleanExpression"]] = None
+
+    @root_validator(skip_on_failure=True)
+    def validate_only_one_boolean_choice_rule_set(cls, values):
+        set_fields = {
+            k for k, v in values.items() if k not in {"Next"} and v is not None
+        }
+        if len(set_fields) != 1:
+            raise ValueError(
+                'Exactly one of "And", "Or", or "Not" must be specified in a BooleanExpression'
+            )
+        return values
+
+
+class BooleanExpressionWithTransitions(BooleanExpression):
     """
     A Boolean Expression is a JSON object which contains a field named "And",
     "Or", or "Not". If the field name is "And" or "Or", the value MUST be an
@@ -156,28 +173,14 @@ class BooleanExpressionChoiceRule(BaseModel):
     of the boolean to which the Choice Rule evaluates.
     """
 
-    And: t.Optional[t.List[ChoiceRule]] = Field(None, min_items=1)
-    Or: t.Optional[t.List[ChoiceRule]] = Field(None, min_items=1)
-    Not: t.Optional[ChoiceRule] = None
     Next: str
-
-    @root_validator(skip_on_failure=True)
-    def validate_only_one_boolean_choice_rule_set(cls, values):
-        set_fields = {
-            k for k, v in values.items() if k not in {"Next"} and v is not None
-        }
-        if len(set_fields) == 0:
-            raise ValueError(
-                'One of "And", "Or", or "Not" must be specified in a BooleanExpression'
-            )
-        return values
 
 
 class ChoiceState(BaseModel):
     Type: te.Literal[StateType.Choice]
     Comment: t.Optional[str] = None
     Choices: t.List[
-        t.Union[DataTestExpressionChoiceRule, BooleanExpressionChoiceRule]
+        t.Union[BooleanExpressionWithTransitions, DataTestExpressionWithTransitions]
     ] = Field(..., min_items=1)
     Default: t.Optional[str]
 
@@ -354,7 +357,7 @@ class FlowDefinition(BaseModel):
         defined_states = set(states.keys())
         referenced_states = {values["StartAt"]}
 
-        for _, state in states.items():
+        for state in states.values():
             if state.Type not in {StateType.Choice, StateType.Fail}:
                 # Ensure that all state's with a Next property have a valid transition
                 next_state = state.Next
@@ -383,7 +386,7 @@ class FlowDefinition(BaseModel):
             )
         if unreferenced_states:
             raise ValueError(
-                f'Flow definition contained unreachable state(s): {",".join(unreferenced_states)}'
+                f'Flow definition contained unreachable state(s): {", ".join(unreferenced_states)}'
             )
 
         return values
--- a/globus_automate_client/models.py
+++ b/globus_automate_client/models.py
@@ -17,13 +17,13 @@ from pydantic import (
 FlowValidationError
 
 
-def enforce_jsonpath(v: t.Optional[str]):
+def enforce_jsonpath(v: t.Optional[str]) -> t.Optional[str]:
     if v is not None and not v.startswith("$."):
         raise ValueError(f'"{v}" must be a JSONPath and start with "$."')
     return v
 
 
-def check_jsonpath_syntax_in_dict(item: t.Optional[dict]):
+def check_jsonpath_syntax_in_dict(item: t.Optional[dict]) -> t.Optional[dict]:
     if item is None:
         return item
 
@@ -50,6 +50,9 @@ class Catchers(BaseModel):
     ErrorEquals: t.List[str] = Field(..., min_items=1)
     Next: str
     ResultPath: t.Optional[str]
+
+    class Config:
+        extra = Extra.forbid
 
     _enforce_jsonpath = validator("ResultPath", allow_reuse=True)(enforce_jsonpath)
 
@@ -146,6 +149,9 @@ class BooleanExpression(BaseModel):
         None, min_items=1
     )
     Not: t.Optional[t.Union[DataTestExpression, "BooleanExpression"]] = None
+
+    class Config:
+        extra = Extra.forbid
 
     @root_validator(skip_on_failure=True)
     def validate_only_one_boolean_choice_rule_set(cls, values):
@@ -328,6 +334,9 @@ class FlowDefinition(BaseModel):
     States: t.Dict[str, FlowState]
     Comment: t.Optional[str]
 
+    class Config:
+        extra = Extra.forbid
+
     @validator("States")
     def validate_states(cls, v):
         if len(v) < 1:
@@ -346,13 +355,13 @@ class FlowDefinition(BaseModel):
 
     @root_validator(skip_on_failure=True)
     def validate_start_at_is_a_state(cls, values):
-        states, start_at = values.get("States"), values.get("StartAt")
-        if states and start_at not in states:
+        states, start_at = values.get("States", {}), values.get("StartAt", "")
+        if start_at not in states:
             raise ValueError(f'StartAt state "{start_at}" is not defined in States')
         return values
 
     @root_validator(skip_on_failure=True)
-    def validate_all_transtion_states_exist(cls, values):
+    def validate_all_transition_states_exist(cls, values):
         states = values.get("States", {})
         defined_states = set(states.keys())
         referenced_states = {values["StartAt"]}
@@ -382,7 +391,7 @@ class FlowDefinition(BaseModel):
         unreferenced_states = defined_states - referenced_states
         if undefined_states:
             raise ValueError(
-                f'Flow definition contained a reference to unknown state(s): {",".join(undefined_states)}'
+                f'Flow definition contained a reference to unknown state(s): {", ".join(undefined_states)}'
             )
         if unreferenced_states:
             raise ValueError(
--- a/globus_automate_client/models.py
+++ b/globus_automate_client/models.py
@@ -14,6 +14,8 @@ from pydantic import (
     validator,
 )
 
+FlowValidationError
+
 
 def enforce_jsonpath(v: t.Optional[str]):
     if v is not None and not v.startswith("$."):
--- a/globus_automate_client/models.py
+++ b/globus_automate_client/models.py
@@ -1,0 +1,387 @@
+import datetime
+import enum
+import typing as t
+
+import typing_extensions as te
+from pydantic import (
+    AnyHttpUrl,
+    BaseModel,
+    Extra,
+    Field,
+    StrictBool,
+    ValidationError as FlowValidationError,
+    root_validator,
+    validator,
+)
+
+
+def enforce_jsonpath(v: t.Optional[str]):
+    if v is not None and not v.startswith("$."):
+        raise ValueError(f'"{v}" must be a JSONPath and start with "$."')
+    return v
+
+
+def check_jsonpath_syntax_in_dict(item: t.Optional[dict]):
+    if item is None:
+        return item
+
+    for k, v in item.items():
+        if k.endswith(".$") and (not isinstance(v, str) or not v.startswith("$.")):
+            raise ValueError(
+                f'Key "{k}" indicates its value will be a JSONPath, however its value "{v}" is not a JSONPath'
+            )
+        elif not k.endswith(".$") and (isinstance(v, str) and v.startswith("$.")):
+            raise ValueError(
+                f'Value "{v}" is a JSONPath, however its key "{k}" does not indicate a JSONPath'
+            )
+        elif isinstance(v, dict):
+            check_jsonpath_syntax_in_dict(v)
+    return item
+
+
+class StateType(str, enum.Enum):
+    Pass = "Pass"
+    Action = "Action"
+    Wait = "Wait"
+    Choice = "Choice"
+    ExpressionEval = "ExpressionEval"
+    Fail = "Fail"
+
+
+class Catchers(BaseModel):
+    ErrorEquals: t.List[str] = Field(..., min_items=1)
+    Next: str
+    ResultPath: t.Optional[str]
+
+    _enforce_jsonpath = validator("ResultPath", allow_reuse=True)(enforce_jsonpath)
+
+
+class ChoiceRule(BaseModel):
+    Variable: str
+    StringEquals: t.Optional[str]
+    StringEqualsPath: t.Optional[str]
+    StringLessThan: t.Optional[str]
+    StringLessThanPath: t.Optional[str]
+    StringGreaterThan: t.Optional[str]
+    StringGreaterThanPath: t.Optional[str]
+    StringLessThanEquals: t.Optional[str]
+    StringLessThanEqualsPath: t.Optional[str]
+    StringGreaterThanEquals: t.Optional[str]
+    StringGreaterThanEqualsPath: t.Optional[str]
+    StringMatches: t.Optional[str]
+    NumericEquals: t.Optional[str]
+    NumericEqualsPath: t.Optional[str]
+    NumericLessThan: t.Optional[str]
+    NumericLessThanPath: t.Optional[str]
+    NumericGreaterThan: t.Optional[str]
+    NumericGreaterThanPath: t.Optional[str]
+    NumericLessThanEquals: t.Optional[str]
+    NumericLessThanEqualsPath: t.Optional[str]
+    NumericGreaterThanEquals: t.Optional[str]
+    NumericGreaterThanEqualsPath: t.Optional[str]
+    BooleanEquals: t.Optional[str]
+    BooleanEqualsPath: t.Optional[str]
+    TimestampEquals: t.Optional[str]
+    TimestampEqualsPath: t.Optional[str]
+    TimestampLessThan: t.Optional[str]
+    TimestampLessThanPath: t.Optional[str]
+    TimestampGreaterThan: t.Optional[str]
+    TimestampGreaterThanPath: t.Optional[str]
+    TimestampLessThanEquals: t.Optional[str]
+    TimestampLessThanEqualsPath: t.Optional[str]
+    TimestampGreaterThanEquals: t.Optional[str]
+    TimestampGreaterThanEqualsPath: t.Optional[str]
+    IsNull: t.Optional[str]
+    IsPresent: t.Optional[str]
+    IsNumeric: t.Optional[str]
+    IsString: t.Optional[str]
+    IsBoolean: t.Optional[str]
+    IsTimestamp: t.Optional[str]
+
+    class Config:
+        extra = Extra.forbid
+
+    _enforce_jsonpath = validator(
+        "Variable",
+        "StringEqualsPath",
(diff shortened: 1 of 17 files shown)


rudyardrichter:
Looking at this, I think I would be in favor of the following: we move this validation tooling into a separate package entirely, making the "base" validation purely what is needed for the actual state machines API, and making the validators neatly subclassable for our more specific case. It doesn't seem there's much of an ecosystem yet for state machines in Python? It would be a pretty huge undertaking to try to be the ones to publish "the python SFN validation library" but we should probably at least have a conversation about where we want to aim in that respect. The "lower" target to aim for, I think, would still be isolating this stuff to very specific modules in the client/SDK and then importing in flows.

globus_automate_client/models.py
```diff
@@ -0,0 +1,389 @@
...
+                if state.Default:
+                    referenced_states.add(state.Default)
+                for choice in state.Choices:
+                    referenced_states.add(choice.Next)
+
+        undefined_states = referenced_states - defined_states
+        unreferenced_states = defined_states - referenced_states
+        if undefined_states:
+            raise ValueError(
+                f'Flow definition contained a reference to unknown state(s): {",".join(undefined_states)}'
```
kurtmckee, line 382: I think `{", ".join(undefined_states)}` will read better (note the space).

tests/flow_validation/test_action_state.py
```diff
@@ -0,0 +1,339 @@
...
+    FlowDefinition(**flow_def)
+
+
+@pytest.mark.parametrize("flow_def", invalid_flow_definitions)
+def test_invalid_flows_fail_validation(flow_def: t.Dict[str, t.Any]):
+    with pytest.raises(FlowValidationError) as ve:
+        FlowDefinition(**flow_def)
+
+    assert ve.type is FlowValidationError
+    # assert False, ve.value.errors()
```
kurtmckee, line 339: It looks like this comment could be removed.

tests/flow_validation/test_choice_state.py
```diff
@@ -0,0 +1,429 @@
...
+            "Type": "Choice",
+            "Comment": "No info",
+            "Choices": [
+                {
+                    "Variable": "$.SomePath.status",
+                    "NOT_AN_OPERATOR": "FAILED",
+                    "Next": "ChoiceAState",
+                },
+            ],
+            "Default": "DefaultState",
```
kurtmckee, line 387: All of these test cases have a `"Default"` key. It would be good to confirm what happens if no default is specified (should be fine) and if the default state doesn't actually exist..

tests/flow_validation/test_example_flows.py
```diff
@@ -0,0 +1,14 @@
+import glob
+import os
+
+import pytest
+import yaml
+
+from globus_automate_client.models import FlowDefinition
+
+
+@pytest.mark.parametrize("filename", glob.glob("./examples/flows/*/definition*"))
```
kurtmckee, line 10: LOVE that you're testing this! :+1:

tests/flow_validation/test_expression_eval_state.py
```diff
@@ -0,0 +1,133 @@
...
+    },
+}
+
+nested_parameters_with_invalid_json_path_key = {
+    "StartAt": "ExpressionEval",
+    "States": {
+        "ExpressionEval": {
+            "Type": "ExpressionEval",
+            "Comment": "No info",
+            "Parameters": {"invalid_json_path_indicator": "$.some_path"},
```
kurtmckee, line 39: I don't actually know for sure...is it possible to specify static parameters? For example, would this be acceptable?

```json
{"valid_key": "valid value"}
```

If that's acceptable, should we bar values that start with `"$."`?

```json
{"interest_calculated": "$.07"}
```

kurtmckee (changes requested):
Great work Uriel!

In addition to my other comments, I strongly recommend that you add roundtrip tests for all valid tests -- make sure that the dicts can be converted to JSON, then load them back through Python's JSON parser and make sure the final dict matches the original.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+            raise ValueError('Either "Next" or "End" are required')
+        return values
+
+
+class ActionState(NextOrEndState):
+    Type: te.Literal[StateType.Action]
+    Comment: t.Optional[str] = None
+    ActionUrl: AnyHttpUrl
+    ActionScope: t.Optional[str] = None
+    Parameters: t.Optional[t.Dict[str, t.Any]] = None
```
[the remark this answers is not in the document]
jimPruyne: Oops, I'm wrong. :( `InputPath` is not allowed on `ExpressionEval` (which is correct in the model for that state). Perhaps there's still some sharing that could be done, but it is less clear to me upon this further review.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+            raise ValueError(
+                f"Only one of these fields may be set: {set_fields.keys()}"
+            )
+        if len(set_fields) == 0:
+            raise ValueError(f"At least one field must be set: {fields}")
+
+        return values
+
+
+FlowState = te.Annotated[
```
jimPruyne, line 318: This is the clever/correct solution for setting up the discriminator on a type which is only used as a `TypeAnnotation` (in the `States` dictionary below). Nice One!

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+        ]
+        if invalid_state_names:
+            raise ValueError(
+                f"The following states have invalid names: {', '.join(invalid_state_names)}"
+            )
+        return v
+
+    @root_validator(skip_on_failure=True)
+    def validate_start_at_is_a_state(cls, values):
+        states, start_at = values.get("States"), values.get("StartAt")
```
jimPruyne, line 349: Could just default `states` to `{}` on the `get` to avoid the check on `states` below.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+        return values
+
+    @root_validator(skip_on_failure=True)
+    def validate_all_transtion_states_exist(cls, values):
+        states = values.get("States", {})
+        defined_states = set(states.keys())
+        referenced_states = {values["StartAt"]}
+
+        for state in states.values():
+            if state.Type not in {StateType.Choice, StateType.Fail}:
```
jimPruyne, line 361: In my prototype, I used a mixin, abstract base class on the various `State` model classes to have them return the referenced state(s) (e.g. the value for `Next`). I think that reads rather cleanly as it puts that logic close to the state definition itself.

globus_automate_client/models.py
jimPruyne: Isn't that about what we'd get if we just did the equivalent of `str(undefined_states)`?

tests/flow_validation/test_example_flows.py
jimPruyne: Yeah, great idea!!

tests/flow_validation/test_action_state.py
```diff
@@ -0,0 +1,326 @@
+import typing as t
+
+import pytest
+
+from globus_automate_client.models import FlowDefinition, FlowValidationError
+
+simple_action_state = {
```
jimPruyne, line 7: Looks like all of these are single step flows, so we probably could just test directly on the state of interest rather than providing a full flow definition (though we'd need separate test cases for the fields in the flow definition and how that then interacts with unreferenced state names, etc.)

tests/flow_validation/test_action_state.py
```diff
@@ -0,0 +1,326 @@
...
+            "Comment": "No info",
+            "ActionUrl": "http://localhost:5000",
+            "Parameters": {},
+            "ResultPath": "$.some_path",
+            "End": True,
+        },
+    },
+}
+
+action_state_with_catcher = {
```
jimPruyne, line 21: I guess these which have a `Next` value need to be complete flows, so there's some opportunity to do what I suggested above, but perhaps not as many as I thought on my first skim through.

tests/flow_validation/test_action_state.py
```diff
@@ -0,0 +1,326 @@
...
+def test_valid_flows_pass_validation(flow_def: t.Dict[str, t.Any]):
+    flow_model = FlowDefinition(**flow_def)
+    flow_def_out = flow_model.dict(exclude_unset=True)
+    assert flow_def_out == flow_def
+
+
+@pytest.mark.parametrize("flow_def", invalid_flow_definitions)
+def test_invalid_flows_fail_validation(flow_def: t.Dict[str, t.Any]):
+    with pytest.raises(FlowValidationError) as ve:
+        FlowDefinition(**flow_def)
```
jimPruyne, line 324: Probably would be nice to capture in the test case what sort of error is expected during the validation.

jimPruyne (changes requested):
This is really, really good. I had pointed at @kurtmckee at a pretty quick and dirty draft of similar that I had done, and he also has some stuff. Between all of them, I think we have a lot of stuff covered including some of the more complicated `Choice` things, maybe a few decent things I did(?) and all the good stuff here.

Key in all this is great test cases, and there are a lot here which is good. I hadn't gotten that far on test cases which is probably fine at this point.

I'm gonna mark this Request changes, and probably Kurt will have to think about strategy for bringing the 3 threads together, though I'm happy to continue to contribute there.

jimPruyne:
> Looking at this, I think I would be in favor of the following: we move this validation tooling into a separate package entirely, making the "base" validation purely what is needed for the actual state machines API, and making the validators neatly subclassable for our more specific case. It doesn't seem there's much of an ecosystem yet for state machines in Python? It would be a pretty huge undertaking to try to be the ones to publish "the python SFN validation library" but we should probably at least have a conversation about where we want to aim in that respect. The "lower" target to aim for, I think, would still be isolating this stuff to very specific modules in the client/SDK and then importing in flows.

I had also considered making this a separate package mostly for ease(?) of import into both our client and server. I'm not _sure_ it is worth the effort as if we can keep the actual validation a single file, we could copy it into the server (I'd keep the client-side the definitive copy) with the intent of ultimately making the Flows service depend on automate client so we could access the validation (and maybe some other stuff?). As mentioned earlier, the hurdle to that is getting the Flows service Globus SDK v3 compliant which is also a really good idea.

Sort of reiterating my comment on the review, we've got multiple threads including branches on both the service and client sides to bring together. But, I _think_ this branch/PR has the most going for it, so is perhaps the best starting point for the merge of all the code, ideas, tests, etc.

globus_automate_client/models.py
```diff
@@ -0,0 +1,387 @@
...
+    TimestampLessThanEquals: t.Optional[str]
+    TimestampLessThanEqualsPath: t.Optional[str]
+    TimestampGreaterThanEquals: t.Optional[str]
+    TimestampGreaterThanEqualsPath: t.Optional[str]
+    IsNull: t.Optional[str]
+    IsPresent: t.Optional[str]
+    IsNumeric: t.Optional[str]
+    IsString: t.Optional[str]
+    IsBoolean: t.Optional[str]
+    IsTimestamp: t.Optional[str]
```
kurtmckee, line 97: These items must be booleans.

kurtmckee (changes requested):

globus_automate_client/models.py
```diff
@@ -0,0 +1,387 @@
...
+    Next: str
+    ResultPath: t.Optional[str]
+
+    _enforce_jsonpath = validator("ResultPath", allow_reuse=True)(enforce_jsonpath)
+
+
+class ChoiceRule(BaseModel):
+    Variable: str
+    StringEquals: t.Optional[str]
+    StringEqualsPath: t.Optional[str]
```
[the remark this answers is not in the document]
urimandujano: That is totally possible but I went the direction of a custom validator for the sake of human-centric errors. Using  `constr` will return an error like `'msg': 'string does not match regex "\\$\\..+"'` whereas the error from custom validation looks [in my opinion] less intimidating: `'"some_value" must be a JSONPath and start with "$."'`.

globus_automate_client/models.py
```diff
@@ -0,0 +1,387 @@
...
+    when the boolean value has been unambiguously determined.
+
+    The value of a Boolean Expression containing a "Not" field MUST be a single
+    Choice Rule, that MUST NOT contain a "Next" field; it returns the inverse
+    of the boolean to which the Choice Rule evaluates.
+    """
+
+    And: t.Optional[t.List[ChoiceRule]] = Field(None, min_items=1)
+    Or: t.Optional[t.List[ChoiceRule]] = Field(None, min_items=1)
+    Not: t.Optional[ChoiceRule] = None
```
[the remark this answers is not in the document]
urimandujano: These choice state are such a pain -- I think I've got this right in my review updates. What I've done is defined base classes "BooleanExpression" and `DataTestExpression` which **do not** support `Next` states and can be used recursively.  `BooleanExpressionWithTransitions` and `DataTestExpressionWithTransitions` subclass these base classes and add support for a `Next` state. Now, the top level `Choice` rule can support choices of either `BooleanExpressionWithTransitions` or `DataTestExpressionWithTransitions` which *should* do the right thing. 

I've added a few tests to demonstrate what does work and what doesn't. Nice job spotting this!

globus_automate_client/models.py
```diff
@@ -0,0 +1,387 @@
...
+
+    @root_validator(skip_on_failure=True)
+    def validate_only_one_boolean_choice_rule_set(cls, values):
+        set_fields = {
+            k for k, v in values.items() if k not in {"Next"} and v is not None
+        }
+        if len(set_fields) == 0:
+            raise ValueError(
+                'One of "And", "Or", or "Not" must be specified in a BooleanExpression'
+            )
```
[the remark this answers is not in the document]
urimandujano: I've updated to handle the first case -- now the check if validating that exactly 1 of the non-Next fields is set to a non-None value.

I'm not sure of how to handle number 2. It's a built in assumption that if the model defines a default value and the input data does not explicitly provide that value, the resulting model instance will contain that default value. If we go ahead and re-serialize a FlowDefinition model it will contain `None` values for all optional/exclusive fields that were unset. There is an option to re-serialize the model by omittitng unset values if there's interest in only getting back fields that were explicitly set. I updated tests with that usage to make sure that the dict that goes in is what comes back out.

globus_automate_client/models.py
```diff
@@ -0,0 +1,389 @@
...
+            raise ValueError(f'StartAt state "{start_at}" is not defined in States')
+        return values
+
+    @root_validator(skip_on_failure=True)
+    def validate_all_transtion_states_exist(cls, values):
+        states = values.get("States", {})
+        defined_states = set(states.keys())
+        referenced_states = {values["StartAt"]}
+
+        for _, state in states.items():
```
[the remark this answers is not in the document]
urimandujano: Good catch, this is better form.

globus_automate_client/models.py
urimandujano: Added!

tests/flow_validation/test_action_state.py
urimandujano: Gone!

tests/flow_validation/test_choice_state.py
urimandujano: Good idea for tests! Added tests for both those cases.

tests/flow_validation/test_expression_eval_state.py
urimandujano: I think you're right that we want to support static parameter values with JSONPath format. I can see someone trying to create/define a JSONPath value such as:

`{"computed_json_path": "$.dynamic_path_to_plugin_elsewhere"}`

I'll get rid of this restriction.

globus_automate_client/models.py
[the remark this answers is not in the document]
kurtmckee: Nice, makes sense!

globus_automate_client/models.py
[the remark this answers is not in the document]
kurtmckee: Thanks Uriel!

globus_automate_client/models.py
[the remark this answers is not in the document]
kurtmckee: I've seen the `exclude_unset` arg and I'm anticipating that will always be used. Is there a way to specify an alternate sentinel value besides `None`? I think `Ellipsis` has special meaning to pydantic, but perhaps it, or an empty tuple, might be a usable sentinel that can't be created by JSON or YAML inputs?

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+        fields = ["InputPath", "Parameters"]
+        set_fields = {f: values.get(f) for f in fields if values.get(f) is not None}
+        if len(set_fields) > 1:
+            raise ValueError(
+                f"Only one of these fields may be set: {set_fields.keys()}"
+            )
+        if len(set_fields) == 0:
+            raise ValueError(f"At least one field must be set: {fields}")
+
+        return values
```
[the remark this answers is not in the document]
rudyardrichter: Reading the later parts, I think I'll retract this for now, and probably PR a generalization of this bit of validation as a follow-up.

rudyardrichter:
> ... As mentioned earlier, the hurdle to that is getting the Flows service Globus SDK v3 compliant which is also a really good idea.

Do we have a sense on how difficult that would be from here? I might be ignorant to some aspects of this, but it seems like the main one is just keeping the nexus client compatible. It doesn't look like there are conflicts.

re.:
> ... if we can keep the actual validation a single file, we could copy it into the server ...
I'm really wary of having such a large and complicated set of validation logic not consolidated to a single importable somewhere. If even innocuous tweaks happened on the forks separately, this is a pretty large amount of code to have to keep in mind when resolving those.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+    BaseModel,
+    Extra,
+    Field,
+    StrictBool,
+    ValidationError as FlowValidationError,
+    root_validator,
+    validator,
+)
+
+FlowValidationError
```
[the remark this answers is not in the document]
urimandujano: I'm importing pydantic's `ValidationError` as `FlowValidationError` to keep the pydantic details low for the end user but our linters were complaining that `FlowValidationError` was imported but never used. This was just for appeasing the linters.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+def enforce_jsonpath(v: t.Optional[str]):
```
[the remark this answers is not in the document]
urimandujano: Yes

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+    if v is not None and not v.startswith("$."):
+        raise ValueError(f'"{v}" must be a JSONPath and start with "$."')
+    return v
+
+
+def check_jsonpath_syntax_in_dict(item: t.Optional[dict]):
```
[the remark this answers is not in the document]
urimandujano: And yes! Updated the return annotations

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+    if item is None:
+        return item
+
+    for k, v in item.items():
+        if k.endswith(".$") and (not isinstance(v, str) or not v.startswith("$.")):
```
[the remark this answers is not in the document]
urimandujano: Is there a way to validate that? I haven't seen that usage come up and I wasn't able to find examples of that in the JSONPath docs.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+    TimestampGreaterThanEqualsPath: t.Optional[str]
+    IsNull: t.Optional[str]
+    IsPresent: t.Optional[str]
+    IsNumeric: t.Optional[str]
+    IsString: t.Optional[str]
+    IsBoolean: t.Optional[str]
+    IsTimestamp: t.Optional[str]
+
+    class Config:
+        extra = Extra.forbid
```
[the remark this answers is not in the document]
urimandujano: On the double check I noticed that I was allowing extra fields at the definition top level, in catcher fields in the action state, and in boolean expressions. I've added the config and tests verifying its behavior.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+    The value of a Boolean Expression containing a "Not" field MUST be a single
+    Choice Rule, that MUST NOT contain a "Next" field; it returns the inverse
+    of the boolean to which the Choice Rule evaluates.
+    """
+
+    Next: str
+
+
+class ChoiceState(BaseModel):
+    Type: te.Literal[StateType.Choice]
```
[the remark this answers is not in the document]
urimandujano: I considered that but decided it was a low value add since `Type`is defined explicitly on each state anyways. Getting the `Comment` field via a mixin seemed like unnecessarily complex.

globus_automate_client/models.py
```diff
@@ -0,0 +1,392 @@
...
+        t.Union[BooleanExpressionWithTransitions, DataTestExpressionWithTransitions]
+    ] = Field(..., min_items=1)
+    Default: t.Optional[str]
+
+    class Config:
+        extra = Extra.forbid
+
+
+class NextOrEndState(BaseModel):
+    Next: t.Optional[str] = None
```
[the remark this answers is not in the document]
urimandujano: Ditto here - the validation for this happens at the top level FlowDefinition in `validate_all_transtion_states_exist`.

globus_automate_client/models.py
urimandujano: Nice!

globus_automate_client/models.py
urimandujano: That's a really interesting take on the problem, definitely an alternate solution!

globus_automate_client/models.py
urimandujano: @jimPruyne for some reason i wasn't able to response directly to your comment about string-ing a list. Looks like python isn't as clever as it pretends to be:

```
>>> l = ["a", "bb", "ccc"]
>>> str(l)
"['a', 'bb', 'ccc']"
```

tests/flow_validation/test_action_state.py
urimandujano: This is really a great idea - it removes complexity/indentation in the test cases and removes a potential source for exceptions that get raised.

tests/flow_validation/test_action_state.py
urimandujano: Actually... since the `Next` values get validated at the `FlowDefinition` level, there is no validation happening at each state's level that ensures that `Next` points to something that exists. So I think your comment above makes sense and would greatly simplify the test cases. I'm going to push a commit with your style recommendations and then another commit where I change the state tests to only test the state-relevant fields.

tests/flow_validation/test_action_state.py
urimandujano: I know there's some preference for validating error messages in tests but I decided against doing that here since the error messages that get returned are up for review and changes in those error messages would mean a lot of changes to the tests.

globus_automate_client/models.py
urimandujano: Good catch! Added a commit that fixes this and adds this to the tests.

globus_automate_client/models.py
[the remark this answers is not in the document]
urimandujano: If the field is not set (or submodel validation failed) pydantic will default the field value to `None` which technically means `InputPath` and `Parameters` are always in values, hence the filtering.

urimandujano:
I think I've addressed most requests for changes and fielded questions. If you're not out of steam yet please take another pass. I think this structure captures 93% of the state languages format -- it probably needs a little work on validating JSONPath syntax but it is doing more validation than we had server side.

My thoughts on validation direction - my hope was that putting it in the client (and enabling by default) would let us know if there were any issues in the validation logic that user flows ran into while being more forgiving (the validation can be turned off/ignored). If it seemed to validate as expected, I thought we could move it into the service itself (where validation is less forgiving). Deduplicating code is worth it but testing/doc-ing/maintaining more repositories takes non-negligible effort. I'm not against having the service be the authority for validation and having client side validation tap into that.

kurtmckee:
@urimandujano Thank you so much for your work on this!

urimandujano:
Glad it was useful! 🙂

PR #140, opened 2022-03-07
Fix tabular `run-list` output

kurtmckee:
The root cause is that `run-enumerate` and `run-list` anticipate different top-level JSON object keys (`"runs"` and `"actions"`, respectively) but the `run-list` command calls the same function as `run-enumerate` if no `--flow-id` is specified.

The mismatch of anticipated top-level keys prevented `run-list` from finding results to present in tabular form.

Notably, changing the URL route from `.../actions` to `.../runs` changes the top-level key name, too.

JSON and YAML output formats were not affected.


kurtmckee:
*Fixed some failing unit tests*

joshbryan-globus (approved):
LGTM

PR #150, opened 2022-04-08
Dependency updates

kurtmckee:
This introduces several dependency updates, [one of which (click)](https://github.com/pallets/click/issues/1752) contains a security fix. There are other items addressed, see the changelog below for details.

Note: Upgrading typer currently prevents us from keeping typer-cli as a development dependency. There is a temporary hack in the Makefile to mitigate this. The issue is being tracked in tiangolo/typer-cli#50.

# Changelog

Development
-----------

-   Update click to version 8.0.4.
    This resolves a security issue.
-   Update typer to version 0.4.1.
-   Update scriv to version 0.14.0.
    scriv is a development dependency.
-   Temporarily remove typer-cli as a listed development dependency.
    It is still needed when generating the CLI documentation.
-   Add safety as a test environment for local and CI testing.
-   Test against Python 3.10 in CI.

### DIFF
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -1,3 +1,4 @@
+import os
 import uuid
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Type, TypeVar, Union
 from urllib.parse import quote, urlparse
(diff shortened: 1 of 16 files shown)


LeiGlobus (approved):

kurtmckee:
@rudyardrichter Could this merge to the upcoming release, or should we target the next release cycle?

kurtmckee:
[sc-14418]

kurtmckee:
After dropping to pdb it became obvious what the problem was: rich's live rendering is continuing to run while the `flow run-list` subcommand is trying to display an input prompt. The prompt gets printed, but is erased immediately by rich.

We'll likely need to make a decision about which library we rely on for console I/O: rich, typer, or click.

joshbryan-globus (approved):

kurtmckee:
I've introduced a more reliable lock-based mechanism to manage STDOUT I/O and consolidated all of the live content rendering management to a single location.

kurtmckee:
I'm trying to figure out why this is generating the URL `"production.flows.automate.globus.org"` on Python 3.10. It has taken a while to reproduce this on my computer, but I finally figured out that if tox _starts_ running using Python 3.10 then it will exhibit this behavior. I'll try to have this fixed tomorrow, now that I can reproduce it.

kurtmckee:
FINALLY. No thread locking, no Python 3.10 / SDK 3.7.0 test errors, and no STDOUT conflicts on output rendering nor when entering auth tokens nor when dropping to pdb.

This PR -- whose purpose was to bump the click version because old versions were using `mktemp` instead of `mkstemp` -- is ready for final review.

globus_automate_client/action_client.py
```diff
@@ -10,11 +11,34 @@
 _ActionClient = TypeVar("_ActionClient", bound="ActionClient")
 
 
+PRODUCTION_ACTIONS_BASE_URL = "https://actions.globus.org"
+
+_ENVIRONMENT_ACTIONS_BASE_URLS = {
+    None: PRODUCTION_ACTIONS_BASE_URL,
+    "prod": PRODUCTION_ACTIONS_BASE_URL,
+    "production": PRODUCTION_ACTIONS_BASE_URL,
```
LeiGlobus, line 19: I guess we'll have to watch for both prod and production!  Without knowing which other services support which, it is safer to have both options.

tests/test_action_client.py
```diff
@@ -48,7 +48,7 @@ def test_trivial_methods(ac, mocked_responses, name, method):
     """Validate the URL used with trivial requests."""
 
     action_id = "bogus"
-    url = f"https://actions.api.globus.org/{action_id}/{name}"
+    url = f"https://actions.globus.org/{action_id}/{name}"
```
LeiGlobus, line 51: Seeing that there's multiple lines changed in this and test_flow_client, I'd consider declaring and using https://actions.globus.org and https://flows.globus.org as a constant base

LeiGlobus (approved):

LeiGlobus:
Nice work Kurt!

kurtmckee:
Thanks Lei!

tests/test_action_client.py
kurtmckee: That's a great idea (especially since I just declared the Actions URL as a constant)! I'll modify these to use the constants before merging.

kurtmckee: I've amended the last commit and force pushed. After CI runs, I'll merge this branch.

Thanks for the quick review, Lei!

PR #156, opened 2022-05-05
Production

joshbryan-globus:
Merge production back to main.


changelog.d/20220429_170559_kurtmckee_update_rich.rst
```diff
@@ -0,0 +1,7 @@
+Development
+-------------
+
+-   Update the rich dependency to 0.12.3.
```
kurtmckee, line 4: This changelog fragment should be integrated into the 0.15.0 release notes. If you haven't pushed 0.15.0 to PyPI then this can be fixed in the `production` branch before release to PyPI.

I should have added it directly when fixing that issue. Sorry!

kurtmckee (changes requested):

changelog.d/20220429_170559_kurtmckee_update_rich.rst
joshbryan-globus: I already did push to pypi, and didn't see this until just now ... not sure why I didn't get a notice about it.  Should we bump to 0.15.1 just to include the changlog?

kurtmckee: For a non-code change like this I'd suggest 0.15.0post1. I'll double-check the syntax on Monday morning.

kurtmckee (approved):

PR #176, opened 2022-08-30
Fix example flow definition: web-option-with-parameters

NickolausDS:
The web-option-with-parameters definition had some invalid, possibly old,
values which would cause the AP to 500 when deployed and run. Removing
these fixes the flow.

### DIFF
--- a/changelog.d/20220830_105958_nick_fix_web_option_with_parameters.rst
+++ b/changelog.d/20220830_105958_nick_fix_web_option_with_parameters.rst
@@ -1,0 +1,7 @@
+.. A new scriv changelog fragment.
+..
+Documentation
+-------------
+
+- Fix example flow definition web-option-with-parameters
+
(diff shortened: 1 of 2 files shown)


jakeglobus:
I'm going to spend some time today looking into what the underlying issue with those parameters are. Thanks for bringing this to our attention Nick!

NickolausDS:
@jakeglobus np! I see there's a failing check due to the changelog. Should I take any action there and add anything to the PR?

jakeglobus:
It is failing because you are missing a changelog fragment. It wants you to run [scriv create](https://scriv.readthedocs.io/en/latest/commands.html#cmd-create) and then fill in some info about the change.

jakeglobus:
I found the commit in the git log where these parameters were removed from the Action provider definition (they are always included now), but I guess we didn't clean up the example. If you push out your changelog fragment, I'll approve this so we can get this merged. Thanks again Nick!

NickolausDS:
I went ahead and added/amended a new scriv, all checks are passing now. Let me know if anything else is out of place!

jakeglobus (approved):

kurtmckee (approved):
Thanks for catching this, Nick!

PR #178, opened 2022-11-03
docs: updates example inputs for Example Flows (to match current Input Schemas)

jbottigliero:
The current examples were missing a few trailing commas, after adding those in, it looks like there were some inconsistencies between the examples and the actual input schemas. It _does_ appear these examples might have been updated with the expectation the underlying Flows move to our new `format: "globus-collection"` inputs... so feel free to close if that is the actual fix here! :v:

### DIFF
--- a/docs/source/authoring_flows.rst
+++ b/docs/source/authoring_flows.rst
@@ -535,14 +535,10 @@ View the `Move flow definition`_ in the Globus web app.
     :caption: Example Input
 
     {
-        "source": {
-          "id": "ddb59aef-6d04-11e5-ba46-22000b92c6ec",
-          "path": "/~/source-directory"
-        },
-        "destination": {
-          "id": "ddb59af0-6d04-11e5-ba46-22000b92c6ec",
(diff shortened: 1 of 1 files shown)


docs/source/authoring_flows.rst
```diff
@@ -535,14 +535,10 @@ View the `Move flow definition`_ in the Globus web app.
...
-          "path": "/~/source-directory"
-        },
-        "destination": {
-          "id": "ddb59af0-6d04-11e5-ba46-22000b92c6ec",
-          "path": "/~/destination-directory"
-        }
+        "source_endpoint_id": "ddb59aef-6d04-11e5-ba46-22000b92c6ec",
+        "source_path": "/~/source-directory",
+        "destination_endpoint_id": "ddb59af0-6d04-11e5-ba46-22000b92c6ec",
+        "destination_path": "/~/destination-directory",
```
kurtmckee, line 541: We do want to advertise the collection format, so I'm going to close this PR but work to fix the issues that you called out.

PR #183, opened 2022-11-15
Add simplified inline flow examples to docs

ada-globus:
# Background
Our production flows are quite complex and could be difficult for a new author to use as a reference when getting started. This adds simplified versions of these flow definitions as inline examples on the “Authoring Flows” page.

For the purposes of this story, I have largely focused on removing the checks that ensure one of the collections is managed (since a non-subscriber user would need to consume their one free deployed flow, we're okay with this as a substitute limitation) and some explicit error handling that provided defaults when the destination does not exist (as a consequence, now allowing the flow definition to present linearly) as well as reordering the steps to make the definitions easier to read sequentially. I highlighted the production flows in a tip above their corresponding definition to draw users' attention to see a more "complete" example.

# Review
For the purposes of review, I've deployed both flows to production and given you all runner roles:

Two-Stage Transfer: https://app.globus.org/flows/18061a3a-fe67-456e-89b7-dcab09e8b8d2
Move (Transfer and Delete): https://app.globus.org/flows/2594d11a-3584-49d1-b2f4-dcb08a02e71c

For the sake of testing these, I have largely focused on the happy path. If you encounter any unexpected errors, please let me know!

### DIFF
--- a/docs/source/authoring_flows.rst
+++ b/docs/source/authoring_flows.rst
@@ -533,9 +533,10 @@ copied to the destination and then deleted from the source.
     :caption: Example Definition
 
     {
+        "Comment": "A Flow for performing a logical 'move' operation by first transferring from a source to a destination and then deleting from the source",
+        "StartAt": "LookupSourcePath",
         "States": {
             "LookupSourcePath": {
-                "Next": "SetSourceInfo",
                 "Type": "Action",
                 "Comment": "Lookup the source path to determine its type (file/dir) to decide if transfer should be recursive",
                 "WaitTime": 172800,
@@ -545,10 +546,10 @@ copied to the destination and then deleted from the source.
                     "path_only": true,
                     "endpoint_id.$": "$.source.id"
                 },
-                "ResultPath": "$.SourcePathInfo"
+                "ResultPath": "$.SourcePathInfo",
+                "Next": "SetSourceInfo"
             },
             "SetSourceInfo": {
-                "Next": "LookupDestinationPath",
                 "Type": "ExpressionEval",
                 "Comment": "Set the recursive flag",
                 "Parameters": {
@@ -556,10 +557,10 @@ copied to the destination and then deleted from the source.
                     "is_recursive.=": "SourcePathInfo.details.DATA[0].is_folder",
                     "source_folder.=": "SourcePathInfo.details.path"
                 },
-                "ResultPath": "$.SourceInfo"
+                "ResultPath": "$.SourceInfo",
+                "Next": "LookupDestinationPath"
             },
             "LookupDestinationPath": {
-                "Next": "SetDestinationInfo",
                 "Type": "Action",
                 "Comment": "Lookup the destination path to determine its type (file/dir)",
                 "WaitTime": 172800,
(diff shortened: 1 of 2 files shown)


docs/source/authoring_flows.rst
```diff
@@ -519,17 +519,190 @@ Example Flows
...
+                    "items.=": "[source.path]",
+                    "label.=": "getattr('delete_label', 'Delete from Source for Move Flow Run with id ' + _context.run_id)",
+                    "recursive.$": "$.SourceInfo.is_recursive",
+                    "endpoint_id.$": "$.source.id"
+                },
+                "ResultPath": "$.DeleteResult",
+            }
+        },
+        "Comment": "A Flow for performing a logical 'move' operation by first transferring from a source to a destination and then deleting from the source",
+        "StartAt": "LookupSourcePath"
```
derek-globus, line 622: These two should be at the top of the flow def for documentation sake.
If these are the reference examples people will use, the first thing they run into being a keyword telling them where to start in a big complex json blob is helpful

docs/source/authoring_flows.rst
```diff
@@ -549,19 +722,291 @@ View the `Move (copy and delete) files using Globus`_ in the Globus web app.
...
+                    },
+                    "DeleteFromIntermediateInput": {
+                        "items.=": "[(intermediate.path + '/' + _context.run_id)]",
+                        "label.=": "getattr('delete_intermediate_label', 'Delete from Intermediate for Flow Run with id ' + _context.run_id)",
+                        "recursive": true,
+                        "endpoint_id.$": "$.intermediate.id"
+                    }
+                },
+                "ResultPath": "$.ActionInputs"
+            },
```
derek-globus, line 835: I'm not convinced this state helps with legibility more than it hurts.

Seems like the idea was to take all of the complexity for the next 4 states & bundle them into one.
Could we just move each `$.ActionInputs.*Input` to the state which uses that input? As best as I can tell we're not re-using any of the data generated here, so it's just de-localizing it.

docs/source/authoring_flows.rst
```diff
@@ -519,17 +519,190 @@ Example Flows
...
 
-View the `Move (copy and delete) files using Globus`_ in the Globus web app.
-(You may need to log in first.)
+.. code-block:: json
+    :caption: Example Definition
+
+    {
+        "States": {
+            "LookupSourcePath": {
+                "Next": "SetSourceInfo",
```
derek-globus, line 538: I kinda think `Next` and `End` tags should go at the end of the state definition rather than the start. My line of thinking would be
1. The reader should learn what the thing is from the first couple attributes attributes (type, actionurl, and comment should adequately cover that)
2. Putting next in the definition will have a direct link (because of the sequential state authoring you did here) to the state 2 lines below meaning the reader isn't ever required to jump around the file

derek-globus (approved):
Left some nitpicky structuring comments.

I tried to be nitpicky around structuring it like the reader hasn't seen a flow before.
If you disagree on the intended reader or any of my notes making it less legible I can be happy as-is getting this into the docs.

ada-globus:
Not nitpicky at all! It sounds like our inclinations are the same here. Those (particularly re: ordering) caught my eye but I had "stetted" them for now because that's how the existing production flows present them when displayed in the Web App. But the fact that you noticed this as well makes me think the revisions would be worthwhile. Will follow up with an update!

derek-globus (approved):

PR #184, opened 2022-12-08
Migrate docs

kurtmckee:
This PR migrates the documentation for

* an overview of the Flows service
* action providers hosted by Globus
* flow authoring

to the Globus documentation site at docs.globus.org. Those changes are hosted in another repository, so most of this will appear as mass deletion of content in this repository.


ada-globus (approved):
This looks good—thanks for thinking to clean up the links in "Using the CLI"! Approving the change now, but we should hold this so that it happens subsequent to the update on the Globus docs site.

kurtmckee:
_Rebased on `main` to prepare for future merge._

sirosen (approved):
Even though we are continuing to refine the docs.globus.org content, I think this change is ready to go.

I like that we're providing links (at least temporarily) to the new docs from old pages.

kurtmckee:
I see that the Globus documentation has been published. I'll merge now, which will trigger an update to ReadTheDocs, and I'll monitor its progress and provide fast-follow PR's if needed.

kurtmckee:
It appears that the Automate CLI docs are linking to the Globus docs as expected. :+1:

PR #185, opened 2022-12-12
Use poetry-core as the build system

kurtmckee:
This allows tools like `tox` to install mandatory dependencies much more quickly (due to only requiring standalone `poetry-core` instead of `poetry` plus all of its many CLI dependencies).


sirosen (approved):
I want to get the builds working again, but don't want us to treat that as a blocker for merging.

I'm starting up on a new branch for the build issues.

PR #186, opened 2022-12-15
Update ci and build configs

sirosen:
Ok. So. Many changes...

- improve GHA workflow to use multiple jobs
- use the check-hook-alias style we worked out for pre-commit to leverage it as the source of truth for versions
- fix the py3.6 builds
- add py3.11 to CI
- make `pre-commit run -a` pass (applies fixers)
- add flake8-bugbear, but ignore B008 due to a conflict with typer
  - fix a bug found by flake8-bugbear
- add dependabot
- add check-jsonschema

The changelog fragments highlight only:
- support for py3.11
- the bugfix


.github/workflows/build.yaml
```diff
@@ -0,0 +1,58 @@
...
+        # we do not want a large number of windows and macos builds, so
+        # enumerate them explicitly
+        include:
+          - os: windows-latest
+            python-version: 3.9
+          - os: macos-latest
+            python-version: 3.9
+          - os: ubuntu-20.04
+            python-version: 3.6
+    name: "test py${{ matrix.python-version }} on ${{ matrix.os }} "
```
kurtmckee, line 20: I prefer more readable test names. Below, my suggestion adds an `os-name` (and removes trailing whitespace in the test name), resulting in test names like "Test py3.11 on Windows".

```suggestion
        os: [ubuntu-latest]
        os-name: [Ubuntu]
        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
        # we do not want a large number of windows and macos builds, so
        # enumerate them explicitly
        include:
          - os: windows-latest
            os-name: Windows
            python-version: 3.9
          - os: macos-latest
            os-name: MacOS
            python-version: 3.9
          - os: ubuntu-20.04
            os-name: Ubuntu 20.04
            python-version: 3.6
    name: "Test py${{ matrix.python-version }} on ${{ matrix.os-name }}"
```

pyproject.toml
```diff
@@ -19,6 +19,11 @@ classifiers = [
     "Development Status :: 4 - Beta",
     "Intended Audience :: Developers",
     "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
kurtmckee, line 26: Poetry [adds these automatically](https://python-poetry.org/docs/pyproject/#classifiers). The Python version classifiers can be completely removed.

```suggestion
```

kurtmckee (changes requested):

pyproject.toml
sirosen: Thank you! I had no idea.

.github/workflows/build.yaml
sirosen: I'm somewhat agnostic towards this change, but I understand why you prefer it. I generally wouldn't bother doing it myself unless pressed, but I more than happily accept it if someone else does it! 😁 

Also, good catch on the trailing space I had in there!

Thanks for the suggestion; applying it now.

kurtmckee (approved):

PR #188, opened 2023-02-07
Support getting the flow definition and schema used to start a run

kurtmckee:
Features
--------

-   Support retrieval of the flow definition and input schema used to start a run.

    This is supported in the new ``flow run-definition`` subcommand.

### DIFF
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -133,7 +133,7 @@ class ActionClient(BaseClient):
 
     def get_definition(self, action_id: str) -> GlobusHTTPResponse:
         """
-        Query the Action Provider for the status of executed Action
+        Get the flow definition for a given run ID.
 
         :param action_id: An identifier that uniquely identifies an Action
             executed on this Action Provider.
--- a/globus_automate_client/action_client.py
+++ b/globus_automate_client/action_client.py
@@ -131,6 +131,16 @@ class ActionClient(BaseClient):
 
         return self.get(f"{quote(action_id)}/status")
 
+    def get_definition(self, action_id: str) -> GlobusHTTPResponse:
+        """
+        Query the Action Provider for the status of executed Action
+
+        :param action_id: An identifier that uniquely identifies an Action
+            executed on this Action Provider.
+        """
+
+        return self.get(f"/runs/{action_id}/definition")
+
     def resume(self, action_id: str) -> GlobusHTTPResponse:
         """
         Resume an INACTIVE action. Corrective action must have been taken prior to invoking
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -898,6 +898,32 @@ def flow_action_status(
     ).run_and_render()
 
 
+@app.command("run-definition")
+def get_flow_definition_for_run(
+    run_id: str = typer.Argument(...),
+    flow_id: uuid.UUID = typer.Option(
+        ...,
+        help="The ID for the Flow which triggered the Action.",
+    ),
+    flow_scope: str = typer.Option(
+        None,
+        help="The scope this Flow uses to authenticate requests.",
+        callback=url_validator_callback,
+    ),
+    flows_endpoint: str = flows_env_var_option,
+    output_format: OutputFormat = output_format_option,
+    verbose: bool = verbosity_option,
+):
+    """
+    Get the flow definition and input schema used to start a run.
+    """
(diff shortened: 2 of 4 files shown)


globus_automate_client/cli/flows.py
```diff
@@ -898,6 +898,32 @@ def flow_action_status(
     ).run_and_render()
 
 
+@app.command("run-definition")
+def get_flow_definition_for_run(
+    run_id: str = typer.Argument(...),
```
ada-globus, line 903: "run" = ❤️

globus_automate_client/action_client.py
```diff
@@ -131,6 +131,16 @@ def status(self, action_id: str) -> GlobusHTTPResponse:
 
         return self.get(f"{quote(action_id)}/status")
 
+    def get_definition(self, action_id: str) -> GlobusHTTPResponse:
+        """
+        Query the Action Provider for the status of executed Action
+
+        :param action_id: An identifier that uniquely identifies an Action
+            executed on this Action Provider.
+        """
```
sirosen, line 140: I recognize that the class hierarchy here is all kinds of misleading, and that this is both a client for APs and-also the client for Specific Flows.

But what should we put in the docstring here? (Right now looks like a copy-paste)

I'm half-tempted to say that the docstring just says "for Flows only" and nothing more?

sirosen (approved):
I admit that I largely was comparing against surrounding code to see if things matched up, but by that standard this seems good.

I noticed one docstring which merits attention.

Since this library and CLI is being deprecated anyway, I'm inclined to allow this to move forward with less stringent review than is typical.

Therefore, I'm pre-approving it right now! If you take a look at the docstring issue and modify only that, please consider this approval valid for whatever reasonable change you make there. 👍

globus_automate_client/cli/flows.py
```diff
@@ -898,6 +898,32 @@ def flow_action_status(
...
+    ),
+    flows_endpoint: str = flows_env_var_option,
+    output_format: OutputFormat = output_format_option,
+    verbose: bool = verbosity_option,
+):
+    """
+    Get the flow definition and input schema used to start a run.
+    """
+    fc = create_flows_client(CLIENT_ID, flows_endpoint)
+    method = functools.partial(
```
ada-globus, line 921: I wouldn't spend more than a handful of minutes on this if it is even possible, but given that the route doesn't require the flow ID nor permissions on the flow itself, do we need to solicit it in this command?

globus_automate_client/cli/flows.py
```diff
@@ -898,6 +898,32 @@ def flow_action_status(
...
+        None,
+        help="The scope this Flow uses to authenticate requests.",
+        callback=url_validator_callback,
+    ),
+    flows_endpoint: str = flows_env_var_option,
+    output_format: OutputFormat = output_format_option,
+    verbose: bool = verbosity_option,
+):
+    """
+    Get the flow definition and input schema used to start a run.
```
ada-globus, line 918: `<pedantic>`

```suggestion
    Get the flow definition and input schema used to start this run.
```

`</pedantic>`

ada-globus (approved):
This seems good to me! I noted a couple minor thoughts, but I'm totally fine if this merges as-is.

globus_automate_client/cli/flows.py
kurtmckee: No, we do not. This same requirement exists on all of the run-related routes, due to how the authorizers are set up (and I'm pretty sure I created a story about removing this requirement, but the priority continues to be to migrate to the Globus CLI).

I already didn't spend more than a handful of minutes on this, which is why this requirement still exists here. 😢😞

globus_automate_client/action_client.py
kurtmckee: You're correct, this was copy-and-paste. It took all of two minutes to create IIRC.

I can add some text that says "for Flows only". I'm leery of spending more than two additional minutes on this PR, though, so heads-up that I hold that opinion!

globus_automate_client/action_client.py
```diff
@@ -131,6 +131,16 @@ def status(self, action_id: str) -> GlobusHTTPResponse:
 
         return self.get(f"{quote(action_id)}/status")
 
+    def get_definition(self, action_id: str) -> GlobusHTTPResponse:
+        """
+        Query the Action Provider for the status of executed Action
```
kurtmckee, line 136:
```suggestion
        Get the flow definition for a given run ID.
```

PR #191, opened 2023-03-14
Prevent passive flow permission updates

kurtmckee:
Bugfixes
--------

-   [\[sc-19291\]](https://app.shortcut.com/globus/story/19291)    Prevent passive flow permission updates.

    Previously, permissions were erased during updates if no permissions were specified.
    Now, permissions will only be erased if an empty string is passed. For example:

    ```text
    # Erase all viewer permissions.
    globus-automate flow update $UUID --flow-viewer=""
    ```

### DIFF
--- a/globus_automate_client/cli/flows.py
+++ b/globus_automate_client/cli/flows.py
@@ -303,51 +303,64 @@ def flow_update(
         "--keyword",
         help="A keyword which may categorize or help discover the Flow. [repeatable]",
     ),
-    flow_viewer: List[str] = typer.Option(
+    flow_viewer: Optional[List[str]] = typer.Option(
         None,
-        help="A principal which may view this Flow. "
-        + _principal_description
-        + "The special value of 'public' may be used to "
-        "indicate that any user can view this Flow. [repeatable]",
-        callback=custom_principal_validator({"public"}),
+        help=(
+            "A principal which may view this flow. "
+            f"{_principal_description} "
+            "\n\nThe special value of 'public' may be used to "
+            "indicate that any user can view this flow. "
+            "\n\nThis option can be used multiple times."
+            "\n\nTo erase any existing viewer permissions, "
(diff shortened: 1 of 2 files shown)


sirosen (approved):
It took me a moment to learn what that callback does, but nothing else even made me hesitate here.

sirosen:
Oh, and the mypy failure should be solvable by just making new variables with `list[str] | None` as their types.

globus_automate_client/cli/flows.py
```diff
@@ -383,6 +396,17 @@ def flow_update(
...
+    # Special cases:
+    # * If the user specifies a single empty string, replace [""] with []
+    #   so all values currently set on the flow will be erased.
+    # * If the user specifies nothing, replace the default empty list with None
+    #   to prevent erasure of the values currently set on the flow.
+    flow_viewer = [] if flow_viewer == [""] else (flow_viewer or None)
+    flow_starter = [] if flow_starter == [""] else (flow_starter or None)
+    flow_administrator = (
+        [] if flow_administrator == [""] else (flow_administrator or None)
+    )
```
derek-globus, line 408: I think it'd be clearer to add a custom flag mutexed with these keys instead of a special value.

Something like
`globus-automate update-flow --no-flow-viewers --no-flow-administrators --flow-starter urn:...`

kurtmckee:
I fixed it by adjusting the type annotations in the function itself. That was sufficient to make mypy happy. :grinning:

globus_automate_client/cli/flows.py
kurtmckee: Run permissions are handled in the same way (the core logic of this code was copied verbatim from the `update_run()` function).

It will take some additional work to introduce that mutex idea, but it can be done. My gut is to follow what exists for run permissions, but let me know if you have a strong desire to do it this way.

PR #199, opened 2023-08-16
Add a deprecation notice to the top of every HTML page

kurtmckee:
This PR introduces a deprecation warning at the top of each HTML page.

It currently renders like this:

image

image


kurtmckee:
The CI lint error should be fixed by #200.

derek-globus (approved):

sirosen (approved):

